PR #99, opened 2022-07-27
Update deprecated qualifiers

mribeirodantas:
This PR updates deprecated qualifiers as pointed out in https://github.com/seqeralabs/nf-training-public/issues/98


mribeirodantas:
Anyone feel free to also work on this PR! I'm not an expert yet on Nextflow so I'm being careful on the updates in order not to break anything 😅

mribeirodantas:
The `exercises` folder within `nf-training` has DSL1 code, but it doesn't seem to me it is being shown/used in the current version of the training. Any idea what is it? Files from an older version of the training?

chriswyatt1:
> The `exercises` folder within `nf-training` has DSL1 code, but it doesn't seem to me it is being shown/used in the current version of the training. Any idea what is it? Files from an older version of the training?

Yes, so the exercise folder used to have all the snippets of code from the ascii docs for a way of a quick look up. We haven't updated this with the dsl2 conversion, and I am not really sure it is necessary, so we could just remove it.

chriswyatt1 (commented):
Great, everything looks good.

ewels:
> Yes, so the exercise folder used to have all the snippets of code from the ascii docs for a way of a quick look up. We haven't updated this with the dsl2 conversion, and I am not really sure it is necessary, so we could just remove it.

If you want, you could go the other way - use only the file snippets here and then use [rich-codex](https://github.com/ewels/rich-codex) to render SVG images of the files in the docs. Still only have a single source to keep up to date then, but get to have them in both places.

Might not be as nice as having the raw code snippets though, in which case my vote would be to kill the directory. Just an idea.. 😉

chriswyatt1:
> > Yes, so the exercise folder used to have all the snippets of code from the ascii docs for a way of a quick look up. We haven't updated this with the dsl2 conversion, and I am not really sure it is necessary, so we could just remove it.
> 
> If you want, you could go the other way - use only the file snippets here and then use [rich-codex](https://github.com/ewels/rich-codex) to render SVG images of the files in the docs. Still only have a single source to keep up to date then, but get to have them in both places.
> 
> Might not be as nice as having the raw code snippets though, in which case my vote would be to kill the directory. Just an idea.. 😉

Yes, that could be a good plan, esp with the mkdocs renewal of the training material. For now I deleted the folder for simplicity.

PR #102, opened 2022-09-08
Update section on Conda / Micromamba

mribeirodantas:
- Update env.yml for conda environment
- Update micromamba section

### DIFF
--- a/asciidocs/containers.adoc
+++ b/asciidocs/containers.adoc
@@ -366,20 +366,26 @@ In this Gitpod environment, conda is already installed.
 
 === Using conda
 
-A Conda environment is defined using a YAML file, which lists the required software packages.
-For example:
+A Conda environment is defined using a YAML file, which lists the required software packages. The first thing you need to do is to initiate conda for shell interaction, and then open a new terminal by running bash.
+
+```
+conda init
+bash
+```
+
+Then write your YAML file. For example:
 
 ```yaml
 name: nf-tutorial
 channels:
+  - conda-forge
   - defaults
   - bioconda
-  - conda-forge
 dependencies:
-  - salmon=1.0.0
-  - fastqc=0.11.5
-  - multiqc=1.5
-  - tbb=2020.2
+  - bioconda::salmon=1.5.1
+  - bioconda::fastqc=0.11.9
+  - bioconda::multiqc=1.12
+  - conda-forge::tbb=2020.2
 ```
 
 Given the recipe file, the environment is created using the command shown below:
@@ -398,8 +404,8 @@ This should look something like this:
 ```bash
 # conda environments:
 #
-base                     /opt/conda
-nf-tutorial           *  /opt/conda/envs/nf-tutorial
+base                  *  /opt/conda
+nf-tutorial              /opt/conda/envs/nf-tutorial
 ```
 
 To enable the environment you can use the `activate` command:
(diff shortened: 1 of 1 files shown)


asciidocs/containers.adoc
```diff
@@ -426,22 +432,38 @@ See the https://www.nextflow.io/docs/latest/conda.html[docs] for details.
...
 
 This saves having to build a conda environment each time you want to use it (as outlined in previous sections).
 
-To do this, you simply require a `Dockerfile`, such as the one in this public https://github.com/nextflow-io/rnaseq-nf/blob/master/docker/Dockerfile-mamba[repo]:
+To do this, you simply require a `Dockerfile` and you use micromamba to install the packages. However, a good practice is to have a YAML recipe file like in the previous section, so we'll do it here too.
 
 ```
-FROM mambaorg/micromamba                                                                                    
-MAINTAINER Your name <your_email>
+name: base
```
ewels, line 442: I think you don't need this? Feels a bit weird to call a new env `base` 🤔 We used to use a custom name which I prefer..

mribeirodantas: I removed it from the YAML file. Is this fine or do you think it should be done differently?

ewels: I think it's probably fine (assuming it still works).

What I was referring to is this - having a [custom environment name](https://github.com/nf-core/methylseq/blob/03972a686bedeb2920803cd575f4d671e9135af0/environment.yml#L3) and then using that when [adding to the `PATH`](https://github.com/nf-core/methylseq/blob/03972a686bedeb2920803cd575f4d671e9135af0/Dockerfile#L16).

Advantage of this is that anyone running these commands not in docker will not pollute / mess up their local conda install. But if it's strictly inside Docker only then it doesn't make much / any difference I think.

mribeirodantas: It still works, and yeah it's strictly inside Docker. It's only used when creating the Docker image. It's not used afterward.

mribeirodantas: It works! :-)

mribeirodantas:
LGTM? 😅

asciidocs/containers.adoc
```diff
@@ -426,22 +432,40 @@ See the https://www.nextflow.io/docs/latest/conda.html[docs] for details.
...
+dependencies:
+  - bioconda::salmon=1.5.1
+  - bioconda::fastqc=0.11.9
+  - bioconda::multiqc=1.12
+  - conda-forge::tbb=2020.2
+```
+
+Then, we can write our Dockerfile with micromamba installing the packages from this recipe file.
+
+```
```
ewels, line 456:
````suggestion
```Dockerfile
````

File: asciidocs/containers.adoc [hunk 3]
```diff
@@ -426,22 +432,40 @@ See the https://www.nextflow.io/docs/latest/conda.html[docs] for details.
...
 
-`micromamba` is fast and robust package for building small conda-based environments.
+`micromamba` is a fast and robust package for building small conda-based environments.
 
 This saves having to build a conda environment each time you want to use it (as outlined in previous sections).
 
-To do this, you simply require a `Dockerfile`, such as the one in this public https://github.com/nextflow-io/rnaseq-nf/blob/master/docker/Dockerfile-mamba[repo]:
+To do this, you simply require a `Dockerfile` and you use micromamba to install the packages. However, a good practice is to have a YAML recipe file like in the previous section, so we'll do it here too.
 
 ```
```
ewels, line 441:
````suggestion
```yaml
````

ewels (approved):
LGTM 👌🏻 

Haven't tested, but reading through it looks fine 👍🏻

christopher-hakkaart (approved):
Looks good

PR #110, opened 2022-09-19
Minor fixes

mribeirodantas:
- Fix minor typo and trailing white spaces in README
- Fix minor typos in the training material
- Replace Ciao by Olá in example with params

### DIFF
--- a/asciidocs/modules.adoc
+++ b/asciidocs/modules.adoc
@@ -487,7 +487,7 @@ The `addParams` option can be used to extend the module parameters without affec
 params.foo = 'Hola'
 params.bar = 'mundo!'
 
-include { SAYHELLO } from './modules.nf' addParams(foo: 'Ciao')
+include { SAYHELLO } from './modules.nf' addParams(foo: 'Olá')
 
 workflow {
     SAYHELLO()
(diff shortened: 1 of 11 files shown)


asciidocs/modules.adoc
```diff
@@ -199,7 +199,7 @@ workflow  {
 }
 ----
 
-If a process defines two or more output channels, each channel can be accessed by indexing the `.out` attribute, e.g., `.out[0]`, `.out[1]`, etc. In our example we only have the `[0]`'th output:
+If a process defines two or more output channels, each channel can be accessed by indexing the `.out` attribute, e.g., `.out[0]`, `.out[1]`, etc. In our example we only have the `[0]'th` output:
```
ewels, line 202: Slightly controversial, but I'm happy if there's a reason (eg. it's easier to read this way)...

ewels (approved):
LGTM!

asciidocs/modules.adoc
mribeirodantas: Before, it rendered like this 😅 
Screen Shot 2022-09-19 at 02 00 20
Now:
Screen Shot 2022-09-19 at 02 01 06

I'm fine if you want to suggest some other way. I just found a similar way that rendered correctly 😄

ewels: Then yeah, that's a good reason :)

PR #112, opened 2022-09-27
Remove trailing white spaces from script files + fix code/output boxes

mribeirodantas:
This Pull Request consists of:

- [x] Remove trailing white spaces from script files (script1-7.nf)
- [x] Fix/standardize source code / output boxes
  - [x] Command lines should be of CMD type, not BASH
  - [x] All code boxes should have the language set, if applicable
  - [x] One line code boxes shouldn’t have line numbers
  - [x] Boxes with 2 or more lines should always have numbered lines (`linenums`)
  - [x] Output from commands shouldn’t be bash. They should be CMD.
  - [x] Standardize the code blocks
Most of the training code blocks consisted of 4 dashes with [keywords,in,this,format]. Some other times, it used ``` or just indentation.
  - [x] There is a snippet in text format, instead of code block (Section 5.2.9)
- [x] Last check for snippets I failed to see at the first time

### DIFF
--- a/asciidocs/awsbatch.adoc
+++ b/asciidocs/awsbatch.adoc
@@ -47,9 +47,10 @@ TIP: The best practice is to keep this setting as a separate
 https://www.nextflow.io/docs/latest/config.html#config-profiles[profile] in your
 workflow config file. This allows the execution with a simple command.
 
-```
+[cmd]
+----
 nextflow run <my pipeline> -profile my-batch-profile
-```
+----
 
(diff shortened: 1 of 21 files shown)


ewels (approved):
Haven't gone through every change, but LGTM from a skim over and quick check of the preview 👍🏻

ewels:
Shame that we don't have syntax highlighting for Nextflow..

ewels:
NB: Will need to switch all of these to `groovy` instead of `nextflow` if we ever implement those.

mribeirodantas:
In this case, it would be nice to find a way to keep the top-right label nextflow + highlight so that it's clear to the reader what is nextflow and what's plain groovy.

PR #114, opened 2022-09-30
New_container_nextflow_22.09.7

chriswyatt1:
New container with latest version of Nextflow 22.09.7edge.
We need to test all the code through the training material before we merge


chriswyatt1:
Container size is still a little big at 3.6GB. Using gitpod/workspace-base image. I did try alpine, but kept getting into failed builds. I can send my attempts if helpful if anyone thinks this is a better route, but this PR will work for now and is half the size

mribeirodantas:
@chriswyatt1 branch `dev` of `nextflow-io/rnaseq-nf` is not working with version `22.09.7-edge`. For now, as the training is already next week, I will replace the example with branch `v2.1` instead. It works both locally and in gitpod if run with `-with-docker`.

mribeirodantas:
@chriswyatt1

```
sudo: singularity: command not found
```

chriswyatt1:
Thank Marcel, ok, good we can sort the `nextflow-io/rnaseq-nf` part. Singularity does not work within gitpod, it is issue #8

abhi18av (approved):
To me, the PR looks good to be merged ✅

PR #119, opened 2022-10-12
SidebarStart

chriswyatt1:
Start sidebar in nf-training directory

### DIFF
--- a/gitpod-ws.code-workspace
+++ b/gitpod-ws.code-workspace
@@ -1,0 +1,5 @@
+{
+    "folders": [
+        { "path": "./nf-training" }
+    ],
+}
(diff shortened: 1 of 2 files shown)


gitpod-ws.code-workspace
```diff
@@ -0,0 +1,43 @@
...
+            "editor.tabSize": 2
+        },
+        "[go]": {
+            "editor.formatOnSave": true
+        },
+        "[tf]": {
+            "editor.insertSpaces": true,
+            "editor.tabSize": 2
+        },
+        "go.formatTool": "goimports",
```
ewels, line 30: Not sure that we need the Go-specific language stuff..

ewels (commented):
Awesome!

gitpod-ws.code-workspace
chriswyatt1: Good point, that should be removed

PR #141, opened 2023-02-02
Keep only Seqera logo in credits

pditommaso:


ewels (approved):
Yup, looks good 👍🏻 

Added a while ago before adding more Nextflow logos into the header and nav etc, so agree that it's no longer needed..

PR #147, opened 2023-02-03
Minor updates

mribeirodantas:
This PR reviews the training material and exercises till the end of module 9.


docs/basic_training/containers.md
```diff
@@ -357,7 +357,7 @@ conda init
 bash
 ```
 
-Then write your YAML file (to `env.yml`). For example:
+Then write your YAML file (to `env.yml`). There is already a file named `env.yml` in the `nf-training` folder as an example. Its content is shown below.
 
 ```yaml
```
ewels, line 362: If it's really identical to the file (no extra comments or anything) then you can just include it directly and it'll always be up to date: https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#embedding-external-files

mribeirodantas: Oh, great suggestion! I will do it, and I believe other places can also benefit from that. Thanks, Phil!

ewels: I think you can do line ranges of the included file as well, but haven't tried that yet.

mribeirodantas:
LGTM @ewels 😄

Issue #148, opened 2023-02-03
Add gitpod buttons


ewels:
Add "launch gitpod" buttons to GitHub readme and docs.

PR #127, opened 2023-01-27
Convert to markdown + mkdocs

ewels:
Converted training material to markdown, using mkdocs material.

Reasoning:

- We want to standardise on using markdown everywhere
- We may soon want to support internationalisation, mkdocs material does this well
- More consistent with other docs
- Better styling, navigation, accessibility
- Timing: May be moving from Seqera to Nextflow branding for the training material

Potentially controversial points:

- I deleted the [AWS docs](https://github.com/seqeralabs/nf-training-public/blob/master/asciidocs/awsbatch.adoc) as these aren't used anywhere
    - About half of the content is duplicated into the executors / deployment scenarios docs
    - The docs are long and super detailed. They're not used anywhere that I can see, which makes them high risk for drifting out of date. IMHO better to delete - they're in the git history if we need them in the future.

TODO:

- [x] Convert asciidoc files to markdown using [this method](https://tinyapps.org/blog/201701240700_convert_asciidoc_to_markdown.html)
- [x] Set up prettier markdown formatting
- [x] Set up mkdocs
- [x] Convert asciidoc admonitions to mkdocs material
- [x] Repeat for nextflow-handson training
- [x] Manually review all markdown files
  - [x] index.md
  - [x] setup.md
  - [x] intro.md
  - [x] rnaseq_pipeline.md
  - [x] containers.md
  - [x] channels.md
  - [x] processes.md
  - [x] operators.md
  - [x] groovy.md
  - [x] modules.md
  - [x] config.md
  - [x] executors.md
  - [x] tower.md
  - [x] cache_and_resume.md
  - [x] debugging.md
- [ ] Set up translations
- [ ] Look into [numbered headings](https://github.com/timvink/mkdocs-enumerate-headings-plugin)
    - [x] Add plugin
    - [ ] Reset numbering for each training (uncomment code in `mkdocs.yml` when https://github.com/timvink/mkdocs-enumerate-headings-plugin/pull/33 merged + released)

PR #149, opened 2023-02-03
Add Run on Gitpod button to README

mribeirodantas:
This is related to issue #148.

What do you have in mind for the docs, @ewels?


ewels:
Nice! Yeah I was thinking something like this, but maybe not in the header - a bit further down?

Originally I was thinking of the button that shows in the automated PRs (eg [here](https://github.com/nextflow-io/training/pull/127#issuecomment-1407200241)). But it's not super pretty so maybe yours is nicer.

Not totally sure for the docs site. But maybe a button on the homepage, also on the setup instructions page?

PR #152, opened 2023-02-03
WIP: Try splitting up long pages

ewels:
⚠️ For feedback only ⚠️ 

I've been wondering if we should split up pages even more, to have bitesize chunks which are a bit easier to digest.
This is a first attempt at playing with it, to see how it feels.

Currently, my gut feeling is that I don't like it very much and we should probably keep things as they are. However, I'd be curious to hear what others think!

👉🏻  Only did up to Channels as it's fairly time consuming, hopefully enough to give a good impression though.


mribeirodantas:
With this change, we have now the left sidebar, plus its expanded version if you click on >, and the right sidebar. I wonder if it's not too much info, in the sense that it may feel confusing for some people.

Apart from that, I don’t really mind. I like that things are very isolated now (concept, example, exercise, bonus), but I’m not sure if it’s worth the work of doing this manually for the whole training material. Anyway, I'm happy both ways, like both versions 😬

ewels:
I like it less and less. Let's kill it.

PR #207, opened 2023-03-08
Small fixes/improvements to the basic training

mribeirodantas:
- [x] Apply small fixes to RNAseq section of basic training (both original and in Portuguese)
- [x] Add how to submit Nextflow as a job to the deployment section of basic training (both original and in Portuguese)
- [x] Improve wording of non-deterministic input channels subsection (cache section)


docs/basic_training/cache_and_resume.md
```diff
@@ -249,63 +249,35 @@ Channel
...
-    """
-}
+```console
+[3, C]
+[4, D]
+[2, B]
+[1, A]
 ```
+
+Imagine now that we have output channels from two processes like this that are input channels to a third downstream process. If this third process expects the ordering to be the same, so that the first element in the first process output channel is related to the first element in the second process output channel, there will be a mismatch. An easy solution for this problem is to use the [fair](https://www.nextflow.io/docs/edge/process.html?highlight=fair#fair) process directive, which applies a fairness concept in multiprocessing that shares resources between the tasks in a fair way so that no task will get more computer resources than the other. Depending on your situation, using the `fair` directive will lead to a decrease in performance. Others have chosen to use tuples assigning an identification to channel elements (`['SRR123', '/some/path/SRR123.fasta']` instead of simply `/some/path/SRR123.fasta`) to make sure the processes are not incurring into a mismatch.
```
ewels, line 283:
```suggestion

..and that order will likely be different every time the pipeline is run.

Imagine now that we two processes like this, whose output channels are acting as input channels to a third process. Both channels will be independently random, so the third process must not expect them to retain a paired sequence. If it does assume that the first element in the first process output channel is related to the first element in the second process output channel, there will be a mismatch.

A common solution for this is to use what is commonly referred to as a _meta map_. A groovy object with sample information is passed out together with the file results within an output channel as a tuple. This can be then be used to pair samples from separate channels together for downstream use. For example, instead of putting just  `/some/path/myoutput.bam` into a channel, you could use `['SRR123', '/some/path/myoutput.bam']` to make sure the processes are not incurring into a mismatch.

If meta maps are not possible, an alternative is to use the [`fair`](https://nextflow.io/docs/edge/process.html#fair) process directive. When specified, Nextflow will guarantee that the order of outputs will match the order of inputs.

!!! warning

     Depending on your situation, using the `fair` directive will lead to a decrease in performance.
```

docs/basic_training/executors.md
```diff
@@ -48,6 +48,36 @@ process {
 }
 ```
 
+### Submit Nextflow as a job
+
+Nextflow should not be run as a command in the login/head node of a cluster, as this node is not prepared for commands that run for too long, even if the amount of resources is negligible. Nextflow should be submitted as a job instead, and in a job node, it will submit new tasks and manage them. In a cluster using Slurm as a job scheduler, for example, you'll need to create a file similar to the one below (save it as `launch_nf.job` or any other name of your choosing):
```
ewels, line 53: I'm not sure I entirely agree with this. I have run Nextflow on the head node for years, I think it very much depends on the cluster setup. Also, many cluster setups do not allow worker nodes to submit jobs, so the proposed solution will often not work.

```suggestion
Whilst the main Nextflow command can be launched on the login / head node of a cluster, be aware that the node must be set up for commands that run for a long time, even if the compute resources used are negligible. Another option  is to submit the main Nextflow process as a job on the cluster instead.

!!! note

    This requires your cluster configuration to allow jobs be launched from worker nodes, as Nextflow will submit new tasks and manage them from here.

For example, if your cluster uses Slurm as a job scheduler, you could create a file similar to the one below:
```

docs/basic_training/executors.md
```diff
@@ -48,6 +48,36 @@ process {
...
+
+```bash linenums="1"
```
ewels, line 55:
```suggestion
```bash linenums="1" title="launch_nf.sh"
```

docs/basic_training/executors.md
```diff
@@ -48,6 +48,36 @@ process {
...
+
+PIPELINE=$1
+CONFIG=$2
+
+conda activate nextflow
+
+nextflow -C ${CONFIG} run ${PIPELINE}
+```
+
+And then run submit it with:
```
ewels, line 70:
```suggestion
And then submit it with:
```

docs/basic_training/executors.md
```diff
@@ -48,6 +48,36 @@ process {
...
+
+```bash linenums="1"
+sbatch launch_nf.job /home/my_user/path/mypipeline.nf /home/my_user/path/myconfig_file.conf
```
ewels, line 73: Using `.sh` gives proper bash syntax highlighting in most code editors.

```suggestion
sbatch launch_nf.sh /home/my_user/path/mypipeline.nf /home/my_user/path/myconfig_file.conf
```

docs/basic_training/executors.md
```diff
@@ -48,6 +48,36 @@ process {
...
+#!/bin/bash
+#SBATCH --partition WORK
+#SBATCH --mem 5G
+#SBATCH -c 1
+#SBATCH -t 12:00:00
+
+PIPELINE=$1
+CONFIG=$2
+
+conda activate nextflow
```
ewels, line 65:
```suggestion
# Use a conda environment where you have installed Nextflow
# (may not be needed if you have installed it in a different way)
conda activate nextflow
```

docs/basic_training/executors.md
```diff
@@ -48,6 +48,36 @@ process {
...
+nextflow -C ${CONFIG} run ${PIPELINE}
+```
+
+And then run submit it with:
+
+```bash linenums="1"
+sbatch launch_nf.job /home/my_user/path/mypipeline.nf /home/my_user/path/myconfig_file.conf
+```
+
+You can find more details about the example above [here](https://lescailab.unipv.it/guides/eos_guide/use_nextflow.html#large-testing-or-production). Besides, make sure to check the following blog posts on best tips for running Nextflow on HPC:
```
ewels, line 76:
```suggestion
You can find more details about the example above [here](https://lescailab.unipv.it/guides/eos_guide/use_nextflow.html#large-testing-or-production).
You can find more tips for running Nextflow on HPC in the following blog posts:
```

ewels (changes requested):
Had a few suggestions for changes on this one. Only did them on the English, sorry 😉 

Great changes though. I've stumbled on the sections you've edited before, I think they're much improved here..

docs/basic_training/cache_and_resume.md
ewels: Could be nice to give a better example of using meta maps here. I think that we should recommend `fair` as a last resort only. It's the lazy option, but it could have quite a negative impact if its use becomes widespread.

docs/basic_training/cache_and_resume.md
```diff
@@ -249,63 +249,35 @@ Channel
...
+    tuple val(task.index), val(x)
 
-    output:
-    tuple val(pair), path('*.bam'), emit: bam_ch
-
-    script:
+  script:
     """
-    your_command --here
+    sleep \$((RANDOM % 3))
```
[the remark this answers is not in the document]
mribeirodantas: 😬

docs/basic_training/cache_and_resume.md
mribeirodantas: Indeed! Can you point me to one? There are tons of examples in nf-core pipelines but if you have an easy and straightforward example at the tip of your tongue that would be great 😬

docs/basic_training/executors.md
mribeirodantas: I have also run Nextflow from the login node 😬 But it's not ideal (of course, you gotta setup worker nodes to allow the submission of jobs and stuff). Anyways, I like more your version. Committing! 😬

docs/basic_training/cache_and_resume.md
mribeirodantas: Would that be a nice example?

```nextflow
// assume the channel below was generated with fromFilePairs
filepairs = [
    SRR493366,
    [/my/data/SRR493366_1.fastq, /my/data/SRR493366_2.fastq]
]
// A meta map would look like
meta_map = [
    [id: 'SRR493366', single_end: false], // meta map
    [/my/data/SRR493366_1.fastq, /my/data/SRR493366_2.fastq]
]
```

PR #216, opened 2023-03-15
Add real JSON module example + minor fixes

mribeirodantas:
- Fix typo in the exercise of the process section of the basic training
- Sync Portuguese translation of the process section of the basic training
- Add real example to channels section on JSON parser as module

### DIFF
--- a/docs/basic_training/channels.md
+++ b/docs/basic_training/channels.md
@@ -666,26 +666,40 @@ See the following Nextflow script:
 ```groovy linenums="1"
 include{ parseJsonFile } from './modules/parsers.nf'
 
-process foo {
+process FOO {
   input:
-  tuple val(meta), path(data_file)
+    tuple val(patient_id), val(feature)
+  output:
+    stdout
 
   """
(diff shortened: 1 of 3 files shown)


docs/basic_training/channels.md
```diff
@@ -666,26 +666,40 @@ See the following Nextflow script:
...
+    | unique
+    | FOO
+    | view
 }
 ```
 
 For this script to work, a module file called `parsers.nf` needs to be created and stored in a modules folder in the current directory.
 
-The `parsers.nf` file should contain the `parseJsonFile` function.
+The `parsers.nf` file should contain the `parseJsonFile` function. For example:
```
ewels, line 692: Why not just have this as part of the first .nf file, have it all together?

ewels (approved):
Nice 👍🏻

docs/basic_training/channels.md
mribeirodantas: I think this example's original goal is to show that you can have auxiliary functions in separate files included.

PR #233, opened 2023-04-17
Fix snippet to match exercise in process section

mribeirodantas:
- Match exercise solution to code mentioned earlier
- Sync Portuguese translation of the basic training (Process section)


ewels (approved):
LGTM

PR #235, opened 2023-04-20
Update Nextflow version to 23.04.1

mribeirodantas:
- [X] Update recommended NXF version for training material
- [X] Fix exception when NXF_VER=22.10.4 must be used for -dsl1
- [X] Sync Portuguese translation of the channels section
- [ ] Update output for 23.04.1

### DIFF
--- a/docs/basic_training/channels.md
+++ b/docs/basic_training/channels.md
@@ -104,14 +104,17 @@ DataflowQueue(queue=[DataflowVariable(value=1), DataflowVariable(value=groovyx.g
 
 We have the value 1 as the single element of our queue channel and a poison pill, which will tell the process that there’s nothing left to be consumed. That’s why we only have one output for the example above, which is 2. Let’s inspect a value channel now.
 
-```groovy linenums="1"
+```groovy linenums="1" title="example3.nf"
 ch1 = Channel.value(1)
 println ch1
 ```
 
 ```console
-$ nextflow run example.nf -dsl1
-...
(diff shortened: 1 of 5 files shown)


docs/basic_training/channels.md
```diff
@@ -85,33 +85,7 @@ When you run the script, it prints only 2, as you can see below:
...
-```
-
-```console
-$ nextflow run example.nf -dsl1
-...
-DataflowVariable(value=1)
-```
-
-There is no poison pill, and that’s why we get a different output with the code below, where `ch2` is turned into a value channel through the `first` operator.
+A process will only instantiate a task when there are elements to be consumed from all the channels provided as input to it. Because `ch1` and `ch2` are queue channels, and the single element of `ch2` has been consumed, no new process instances will be launched, even if there are other elements to be consumed in `ch1`. To consume multiple times the single element in `ch2` we can either use `Channel.value` as mentioned above, or a channel operator that returns a single element, such as `first` below:
```
ewels, line 88:
```suggestion
A process will only instantiate a task when there are elements to be consumed from _all_ the channels provided as input to it. Because `ch1` and `ch2` are queue channels, and the single element of `ch2` has been consumed, no new process instances will be launched, even if there are other elements to be consumed in `ch1`.

To use the single element in `ch2` multiple times, we can either use `Channel.value` as mentioned above, or use a channel operator that returns a single element such as `first()` below:
```

ewels (dismissed):
Great! Even clearer _without_ the example actually!

ewels (approved):

PR #248, opened 2023-06-29
Improve description of the fair process directive

mribeirodantas:
- Improve mention of the fair directive in the cache subsection
- Update Portuguese and French translation of the cache section

### DIFF
--- a/docs/basic_training/cache_and_resume.md
+++ b/docs/basic_training/cache_and_resume.md
@@ -320,7 +320,7 @@ bam
     | PROCESS_C
 ```
 
-If meta maps are not possible, an alternative is to use the [`fair`](https://nextflow.io/docs/edge/process.html#fair) process directive. When this directive is specified, Nextflow will guarantee that the order of outputs will match the order of inputs. It's important to mention that the order in which the tasks will be finished will not necessarily follow the order in the input channel, but Nextflow guarantees that, by the end of it, the output channel will contain the elements in the respective order.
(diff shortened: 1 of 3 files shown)


docs/basic_training/cache_and_resume.md
```diff
@@ -320,7 +320,7 @@ bam
     | PROCESS_C
 ```
 
-If meta maps are not possible, an alternative is to use the [`fair`](https://nextflow.io/docs/edge/process.html#fair) process directive. When specified, Nextflow will guarantee that the order of outputs will match the order of inputs.
+If meta maps are not possible, an alternative is to use the [`fair`](https://nextflow.io/docs/edge/process.html#fair) process directive. When this directive is specified, Nextflow will guarantee that the order of outputs will match the order of inputs. It's important to mention that the order in which the tasks will be finished will not necessarily follow the order in the input channel, but Nextflow guarantees that, by the end of it, the output channel will contain the elements in the respective order.
```
ewels, line 323:
```suggestion
If meta maps are not possible, an alternative is to use the [`fair`](https://nextflow.io/docs/edge/process.html#fair) process directive. When this directive is specified, Nextflow will guarantee that the order of outputs will match the order of inputs (not the order in which the tasks run, only the order of the output channel).
```

ewels (dismissed):
Great! Maybe a little verbose, suggested change which is a bit shorter. Happy either way.

PR #251, opened 2023-07-31
Clarified Explanation in the "Warning" Section of the Process Page

dxu104:
In the first Warning section of the "process" page in the "basic train", I made changes to the statement: "**Since Nextflow uses the same Bash syntax for variable substitutions in strings, Bash environment variables need to be escaped using the \ character.**" 
I felt the explanation wasn't very clear, so I added the following content: "**When you use a \, the $PWD is a directory that Nextflow creates for that specific process (e.g. work/7f/f285b80022d9f61e82cd7f90436aa4/).**"

### DIFF
--- a/docs/basic_training/processes.md
+++ b/docs/basic_training/processes.md
@@ -132,7 +132,7 @@ workflow {
 
 !!! warning
 
-    Since Nextflow uses the same Bash syntax for variable substitutions in strings, Bash environment variables need to be escaped using the `\` character. When you use a `\`, the current directory is a directory that Nextflow creates for that specific process (e.g. work/7f/f285b80022d9f61e82cd7f90436aa4/). 
+    Since Nextflow uses the same Bash syntax for variable substitutions in strings, Bash environment variables need to be escaped using the `\` character. When you use a `\`, the $PWD is a directory that Nextflow creates for that specific process (e.g. work/7f/f285b80022d9f61e82cd7f90436aa4/). 
 
 ```groovy linenums="1"
 process FOO {
--- a/docs/basic_training/processes.md
+++ b/docs/basic_training/processes.md
@@ -132,7 +132,7 @@ workflow {
 
 !!! warning
 
-    Since Nextflow uses the same Bash syntax for variable substitutions in strings, Bash environment variables need to be escaped using the `\` character.
+    Since Nextflow uses the same Bash syntax for variable substitutions in strings, Bash environment variables need to be escaped using the `\` character. When you use a `\`, the working directory you see is the temporary directory that Nextflow creates for that specific process. However, when you don't use a `\`, the directory you see is the location from which the script was executed.
 
 ```groovy linenums="1"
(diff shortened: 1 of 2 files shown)


mribeirodantas:
What do you think, @dxu104 ?

dxu104:
I believe that in Nextflow scripts, understanding the difference between `\$PWD` and `$PWD` is crucial, as `$PWD` is not just a variable but also an environment variable. In Nextflow scripts, if `$PWD` is not escaped (i.e., not prefixed with a backslash), then Nextflow will attempt to interpret it within its context. If such a variable or environment variable exists in Nextflow's global context, it will use that value. Thus, when you run a Nextflow script, `$PWD` will be resolved to the current directory where you're running the Nextflow script. That's why `$PWD` returns `/Users/XXX/Nextflow_Training/training/nf-training`, which is the folder where the Nextflow subprocess was initiated. `$PWD` is an environment variable automatically set by Unix-like operating systems (including Linux and macOS) in the shell. When you open a terminal or start a new shell session, `$PWD` is set to the current working directory. When you run Nextflow or any other program, the program inherits the environment of its parent process (usually your shell). This means that all environment variables set in the parent process, such as `$PWD`, will be available in the child process (like Nextflow).

However, when you escape `$PWD` (i.e., `\$PWD`), Nextflow won't try to interpret it and will pass it directly to bash. Now, bash will interpret `\$PWD` as the current working directory of the current bash shell. In Nextflow, each task is executed in a separate working directory, which is a subdirectory in Nextflow's `work` directory. That's why the escaped `\$PWD` returns a path like `/Users/XXX/Nextflow_Training/training/nf-training/work/e5/db835a3cf2189ec93bff056d5ff94a`.

Additionally, in the subsequent training part, there's also the use of `\$RANDOM`. It's essential to note the difference between `$RANDOM` and `$PWD`. `$RANDOM` is an internal variable in bash, which generates a random number every time it's referenced. Since `$RANDOM` is a specific feature of bash, a backslash is used to prevent Nextflow itself from trying to interpret it. Without the backslash, an error would occur, because there's no `$RANDOM` variable or environment variable in Nextflow's global context. But `$PWD` won't produce an error, which can be confusing for many beginners.

 @mribeirodantas

mribeirodantas:
I'm sorry, I meant what do you think of the commits I did after yours 😅

Good to merge or you want to change something?

mribeirodantas:
Yes, you're already a contributor. I will merge this Pull Request and with that, you'll be a contributor to the main branch 🥳

mribeirodantas (approved):

mribeirodantas:
You can find yourself in the list of contributors here: https://github.com/nextflow-io/training/graphs/contributors 😉

dxu104:
Highly appreciate. More than happy. This is the first time I contribute for the open source community!

mribeirodantas:
Nice to hear that! Welcome to a whole new world 🤩

PR #252, opened 2023-08-02
A more consistent code style

dxu104:
Instead of **< sample1.fq.gz zcat > sample1.fq**, **using zcat sample1.fq.gz > sample1.fq** is a more common and intuitive method to decompress a gzip file. This approach is more standard and easier to understand.

Using < sample1.fq.gz zcat > sample1.fq is unnecessary and could lead to confusion. In fact, this usage might not work properly in certain cases, as not all versions of zcat support reading from standard input.

### DIFF
--- a/docs/basic_training/processes.md
+++ b/docs/basic_training/processes.md
@@ -916,8 +916,8 @@ process FOO {
 
     script:
     """
-    < sample1.fq.gz zcat > sample1.fq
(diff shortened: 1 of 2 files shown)


mribeirodantas:
Good one, @dxu104 !

mribeirodantas (approved):

PR #254, opened 2023-08-06
more clear explanation

dxu104:
I propose adding a line to clarify this as follows:

In the output, items from the 'left' channel always precede the corresponding items from the 'right' channel."

This would make it clearer for new users and provide a more complete understanding of the function's behavior. Please let me know if this change would be acceptable.

### DIFF
--- a/docs/basic_training/operators.md
+++ b/docs/basic_training/operators.md
@@ -220,7 +220,7 @@ This operator is useful to process a group together with all the elements that s
 
 ### `join()`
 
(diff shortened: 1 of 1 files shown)


mribeirodantas:
I'd rather have the documentation be more complete than the training material than the opposite. This way, I think this contribution would be better placed in the [docs](https://www.nextflow.io/docs/edge/operator.html#join). The `concat` operator has an equivalent comment, so I agree the `join` operator deserves the same. Could you please open this PR there?

dxu104:
> mribeirodantas Hi Marcel, Thank you for your suggestion. I already sent a PR in the nextflow repo.

PR #259, opened 2023-08-17
Update note on Hands-on training

mribeirodantas:
- Update note on hands-on training
- Update pt/es/fr translation of note in hands-on training

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -38,7 +38,7 @@ Find the one that's right for you:
 
     !!! quote inline end ""
 
-        :material-alert: This material is not taught / updated as frequently as the other courses.
+        :material-alert: This material is not taught as frequently as the other courses, but if you're looking for a more hands-on approach with very little theory, this is the right course for you.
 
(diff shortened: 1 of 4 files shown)


docs/index.md
```diff
@@ -38,7 +38,7 @@ Find the one that's right for you:
 
     !!! quote inline end ""
 
-        :material-alert: This material is not taught / updated as frequently as the other courses.
+        :material-alert: This material is not taught as frequently as the other courses, but if you're looking for a more hands-on approach with very little theory, this is the right course for you.
```
ewels, line 41:
```suggestion
        :material-run-fast: This course is quite short and hands-on, great if you're looking for a Nextflow refresher.
```

ewels (changes requested):
I'd argue that it may now be taught as frequently! Reworded to remove that and point out that it's shorter (which isn't mentioned anywhere else).

It might actually be good to write the duration of the courses on the homepage (eg. "Typically 3x 2.5 hours"). But I'll leave that to another PR / to be discussed.

Needs updating in the other translations 🙏🏻

PR #275, opened 2023-09-21
Fix issue on process rnaseq_call_variants not executed at sample level in hands-on training materials

yichao-cai:
Hi,

In the hands-on training, the process `rnaseq_call_variants` is meant to be executed at the sample level (as documented in problem 9 https://training.nextflow.io/hands_on/04_implementation/#process-5-gatk-variant-calling), but instead executed at the replicate level. 

This results in 2 `final.vcf` files produced in 1 sample, and 1 of vcf would be selected and used in the final step (ASE_knownSNPs process).

This error can be spotted in the expected output section (https://training.nextflow.io/hands_on/04_implementation/#bonus-step), where `rnaseq_call_variants ` is executed 6 times but not 3 times.

A fix for this is to do a `groupTuble` on the `rnaseq_gatk_recalibrate.out` channel and before feeding it to `rnaseq_call_variants`

### DIFF
--- a/docs/hands_on/04_implementation.md
+++ b/docs/hands_on/04_implementation.md
@@ -1106,7 +1106,7 @@ The next process has the following structure:
 
     ??? solution
 
-        ```groovy linenums="1" hl_lines="61-70"
+        ```groovy linenums="1" hl_lines="62-71"
         /*
          * Process 5: GATK Variant Calling
          */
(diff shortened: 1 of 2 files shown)


mribeirodantas:
Thanks, @yichao-cai ! I will have a look.

mribeirodantas (approved):

PR #289, opened 2023-10-31
Update .gitpod.yml to remove nf-core tools dev

christopher-hakkaart:


mribeirodantas (approved):
This was a temporary fix that is no longer needed, right?

christopher-hakkaart:
Yup. New version is stable fix

PR #294, opened 2023-12-11
Extend the 'Your first script' section.

ewels:
* Bit more verbose description of what the script is actually doing
* Mini subsection switching the (often unfamiliar) bash commands with some Python

### DIFF
--- a/docs/basic_training/intro.md
+++ b/docs/basic_training/intro.md
@@ -1,5 +1,6 @@
 ---
 description: Getting started with Nextflow
+toc_depth: 1
 ---
 
 # Introduction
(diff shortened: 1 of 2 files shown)


docs/basic_training/intro.md
```diff
@@ -142,7 +143,24 @@ workflow { // (18)!
 21. The final output (in the `results_ch` channel) is printed to screen using the `view` operator (appended onto the channel name).
 22. End of the workflow scope.
 
-The use of the operator `.flatten()` here is to split the two files into two separate items to be put through the next process (else they would be treated as a single element).
+This pipeline takes `params.greeting`, which defaults to the string `Hello world!`, and splits it into indiviudual words in the `SPLITLETTERS` process. Each word is written to a separate file, named `chunk_1`, `chunk_2` and so on. These files are picked up as the process `output`.
```
christopher-hakkaart, line 146:
```suggestion
This pipeline takes `params.greeting`, which defaults to the string `Hello world!`, and splits it into indiviudual words in the `SPLITLETTERS` process. Each word is written to a separate file, named `chunk_aa`, `chunk_ab`,  `chunk_ac`and so on. These files are picked up as the process `output`.
```

christopher-hakkaart (approved):
Small edit but otherwise looks great

docs/basic_training/intro.md
ewels: Ah oops, good spot 👍🏻 

Do you think I should make the Python version have identical behaviour?

ewels: Meh, can't think of a nice syntax to do it that doesn't complicate the example more, so will leave the Python as it is I think.

docs/basic_training/intro.md
```diff
@@ -1,5 +1,6 @@
 ---
 description: Getting started with Nextflow
+toc_depth: 1
```
ewels, line 3:
```suggestion
```

docs/basic_training/intro.md
christopher-hakkaart: Keep it simple - I'd leave it as is

ewels:
Ok, there's something a bit broken with the current versions of mamba + conda somehow. Switching out mamba back to conda in the Dockerfile seems to fix the build. Mamba is still installed, so hopefully it shouldn't affect usage inside the container (as we have a bunch of written training material talking about using it).

ewels:
@mribeirodantas - note some small changes here that are not ported over into the translations. Sorry about that! Shout if I should mark this somewhere.

PR #300, opened 2024-01-18
Draft PR training

christopher-hakkaart:
Hi @lescai 

Overall it looks good.

I took your PR and made a new branch because I was doing some heavy repo modifications and I wasn't sure how it would work out.

I've been through the content and it looks good. You will notice some significant movement in the repo structure to accommodate your work. I added some stylistic changes (changes happening to all training on the site, e.g., explicit exercise blocks when we want the training participant to do something, block headings, and line numbers). I also added more information about how to open GitPod and how to get to the directory that we want them to be in as we can't assume a participant will know where to look for your material.

The only issue that needs to be resolved is the size of the fq and reference files.  I kept them out of the repo for now because they more than double the repo size. Is there any way we can reduce these down while maintaining the real-life applicability? E.g., removing reads and/or fudging some of the reference files? Having these in the repo will load them by default when GitPod starts and I would like to keep it as light as possible so we don't have long wait times when building and environment. 

Cc @mribeirodantas as I moved and updated the RNA-seq material you wrote as well


lescai:
Hi @christopher-hakkaart that's wonderful thanks for all the hard work you've done to accommodate the material within the new repo structure.
About the inputs and reference files: these are also available through my teaching repos, and initially I did add a git clone to the gitpod startup script. One should consider how long it would take for the user to do that.
The reference is located here https://github.com/lescai-teaching/datasets_reference_only (sequence and gatk bundle), and I believe it is also in the test data for Sarek (so might be available in the new test data location for nf-core as well).
The reads have been simulated also for my teaching activities and are located here: https://github.com/lescai-teaching/dataset_exercise_resequencing
Let me know how you would prefer to organise those, and I'd be happy to move/transfer as needed.

vdauwera:
Decision was made to decline these changes iirc

PR #307, opened 2024-02-24
Fixing minor errors

FloWuenne:
This PR fixes three very small erroneous references:
- Welcome: Changed Nextflow tower to Seqera Platform
- Opeartors: Fixed missreferencing of `splitText()` and added `splitCsv()` header
- Processes: Fixed wrong referencing to snippet above when it was the snippet below

Two additional things:
1) `publishDir` gets asked for in an exercise before its introduced. Is this wanted?

2) The gitpod for the training doesn't have singularity, but in general, singularity can work in Gitpod. This [statement](https://github.com/nextflow-io/training/blob/2cd5a8c32919cf6e4bd4f198e378cc57417bac3e/docs/basic_training/containers.md?plain=1#L333) in Dependencies and containers sections makes it sound as if singularity can't work in Gitpod at all. Any reason singularity is not included in the Gitpod image? It is part of the `nfcore/gitpod:dev` image used in nf-core/modules.

### DIFF
--- a/docs/basic_training/index.md
+++ b/docs/basic_training/index.md
@@ -44,6 +44,6 @@ To get you started with Nextflow as quickly as possible, you will be walked thro
 2. Explore Nextflow concepts using some basic workflows, including a multi-step RNA-Seq analysis
 3. Build and use Docker containers to encapsulate all workflow dependencies
 4. Dive deeper into the core Nextflow syntax, including Channels, Processes, and Operators
-5. Cover cluster and cloud deployment scenarios and explore Nextflow Tower capabilities
+5. Cover cluster and cloud deployment scenarios and explore Seqera Platform's capabilities
 
 This will give you a broad understanding of Nextflow, to start writing your own workflows. We hope you enjoy the course! This is an ever-evolving document - feedback is always welcome.
--- a/docs/basic_training/operators.md
+++ b/docs/basic_training/operators.md
@@ -325,6 +325,8 @@ result.large.view { "$it is large" }
 
 ## Text files
 
+### `splitText()`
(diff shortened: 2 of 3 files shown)


FloWuenne:
@christopher-hakkaart

mribeirodantas (approved):

mribeirodantas:
Thanks for the PR, @FloWuenne !

1. Ideally, we'd want to have provided enough information for the reader before asking him to do something with `publishDir`. Are you referring to the second exercise in the _Perform expression quantification_ subsection?
2. We ran into this (or some other) warning and simply didn't insist on making it work. I did some tests now, and it seems you're correct. Do you want to contribute this to the repo? Otherwise I can work on that.

FloWuenne:
Thanks for merging @mribeirodantas ! 😊 

1. Yes exactly. Here the user is asked to add a `publishDir` directive without publishDir ever been mentioned before. If the exercise is for the user to follow the link, figure out what `publishDir` is and then implement it, then its perfect. 
2. I can take care of that. Just to make sure I include everything needed:
 2.1 Add `singularity` to the gitpod docker container here: https://github.com/nextflow-io/training/blob/master/.github/gitpod.Dockerfile
 2.2 Update the section about singularity to remove the 'not working in gitpod' statement and check that all the code presented works with singularity in that container. 
 
Anything else to add for singularity?

mribeirodantas:
I think that's all for point 2. For point 1, I think the link to the documentation on `publishDir` is OK for now, though far from ideal. We're working on something related to the training platform, so I think it should be OK to keep it this way for now 😉

PR #308, opened 2024-02-26
Added singularity to Gitpod container build

FloWuenne:
As discussed in #307 with @mribeirodantas, this PR adds:
- singularity / apptainer to the gitpod container so that participants can use singularity / apptainer instead of docker to do the training
- Removes the warning that singularity does not work in Gitpod: https://github.com/nextflow-io/training/blob/5b1112fedcb2d97d3e8b20755ac114a6c528b94f/docs/basic_training/containers.md?plain=1#L331

I have tested all of the singularity commands given in the training material (including running the nextflow/rnaseq-nf workflow with singularity) and they all worked within the updated container.

### DIFF
--- a/.github/gitpod.Dockerfile
+++ b/.github/gitpod.Dockerfile
@@ -17,6 +17,13 @@ RUN apt-get update --quiet && \
         graphviz \
         software-properties-common
 
+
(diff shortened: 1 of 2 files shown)


mribeirodantas (approved):

mribeirodantas:
Thanks for the contribution, @FloWuenne ! Very nice!

mribeirodantas (approved):

PR #312, opened 2024-03-05
Apply minor changes to improve RNAseq section

mribeirodantas:
When conducting the training in the previous two sessions, I felt these points felt out of place or kind of already giving out the answer for the exercises (exercises that are already quite easy to begin with 😆).


christopher-hakkaart (approved):
LGTM

PR #313, opened 2024-03-12
Gitpod: use latest edge release

ewels:
For the colourz 🌈


ewels:
@mribeirodantas will test that the edge release doesn't break anything before merging.

mribeirodantas (approved):
Tested a few sections of the training material and everything seems to work fine. I won't change the pipeline outputs/version suggested for the training for now, but users are free to run it with the default version which is the one this PR introduces 😉

PR #316, opened 2024-03-20
Advanced update - new operators and bug fixes

robsyme:
- Added detailed `flatMap` and `collectFile` tutorials into the operator tour.
- Fixed subtle `groupKey` bug in the Grouping+Splitting chapter
- Moved all advanced nf and data into dedicated directory to prevent any accidental modification of materials between fundamental and advanced trainings.


docs/advanced/grouping.md
```diff
@@ -302,7 +320,79 @@ MapReads( samples, reference )
 | map { meta, bam -> [meta.subMap('id', 'type'), bam] }
 | combine( intervals )
 | GenotypeOnInterval
+| map { groupKey, bamfile -> [groupKey.target, bamfile] }
```
maxulysse, line 323: Not sure I understand that one

maxulysse (approved):
Looking good

docs/advanced/configuration.md
```diff
@@ -39,7 +39,7 @@ These configuration values would be inherited by every run on that system withou
 Move into the chapter example directory:
 
 ```
-cd advanced/configuration
```
christopher-hakkaart, line 42: Ahhh! Sorry, this was my fault. I moved all of these recently to improve the GitPod structure!

docs/advanced/grouping.md
```diff
@@ -291,6 +291,24 @@ process MergeGenotyped {
...
+}
+| groupTuple
+| CombineBams
+| map { meta, bam -> [meta.subMap('id', 'type'), bam] }
+| combine( intervals )
+| GenotypeOnInterval
+| view { meta, bamfile -> "Meta is of ${meta.getClass()}" }
+```
+
+To remove the `groupKey` and go back to the standard map, we can just call its `target` property. This is now appropriate to send through to the `groupTuple` operator and we will be grouping now only on the elements in the map.
```
adamrtalbot, line 310: Agree with @maxulysse below, it's not very clear. I've rephrased it slightly to say what it is, then how you would use it. Just a suggestion, it might give you some inspiration.

```suggestion
The `groupKey` has a property `target`, which contains the original map. If we call this property, we can replace the groupKey with the original map and get back to the original data structure. This is now appropriate to send through to the `groupTuple` operator and we will be grouping now only on the elements in the map.
```

docs/advanced/grouping.md
```diff
@@ -302,7 +320,79 @@ MapReads( samples, reference )
...
+```output title="Final channel output"
+[[id:sampleB, type:normal], merged.genotyped.bam]
+[[id:sampleB, type:tumor], merged.genotyped.bam]
+[[id:sampleA, type:normal], merged.genotyped.bam]
+[[id:sampleA, type:tumor], merged.genotyped.bam]
+[[id:sampleC, type:normal], merged.genotyped.bam]
+[[id:sampleC, type:tumor], merged.genotyped.bam]
+```
+
+If we would like to save the output of our `MergeGenotyped` process into a save location, we can "publish" the outputs of a process using the `publishDir` directive. We will cover this in more detail on day 2, but try modifying the `MergeGenotyped` process to include the directive:
```
adamrtalbot, line 342:
```suggestion
If we would like to save the output of our `MergeGenotyped` process, we can "publish" the outputs of a process using the `publishDir` directive. We will cover this in more detail on day 2, but try modifying the `MergeGenotyped` process to include the directive:
```

docs/advanced/operators.md
```diff
@@ -455,3 +455,357 @@ Launching `./main.nf` [elegant_rutherford] DSL2 - revision: 2c5476b133
...
+    [sample2, [sample2/data.1.dat, sample2/data.3.dat, sample2/data.2.dat]]
+    [sample2, [sample2/data.4.dat]]
+    ```
+
+    This is a challenging problem that pushes slightly beyond what we have covered.
+
+    !!! tip
+        You may find it helpful to use Groovy's `collate()` method ([docs](http://docs.groovy-lang.org/docs/groovy-2.3.5/html/groovy-jdk/java/util/List.html#collate(int)), [tutorial](https://blog.mrhaki.com/2012/04/groovy-goodness-collate-list-into-sub.html)). This method segments a list into sub-lists of specified size.
+
+        You may also need the `collect()` method, which is like the `map` operator for collections in Groovy ([docs](http://docs.groovy-lang.org/2.4.3/html/groovy-jdk/java/util/Collection.html#collect(groovy.lang.Closure)), [tutorial](https://www.baeldung.com/groovy-lists#Collecting)).
```
adamrtalbot, line 541: Not sure I like the simile here. I think you should explain what `collect` does, then draw a comparison afterwards.

```suggestion
        You may also need the `collect()` method, which will perform an operation on every element in the collection (array, tuple, etc) and return a new collection of the outputs. This is similar to the `map` operator for but works on collections instead of channels ([docs](http://docs.groovy-lang.org/2.4.3/html/groovy-jdk/java/util/Collection.html#collect(groovy.lang.Closure)), [tutorial](https://www.baeldung.com/groovy-lists#Collecting)).
```

docs/advanced/operators.md
```diff
@@ -455,3 +455,357 @@ Launching `./main.nf` [elegant_rutherford] DSL2 - revision: 2c5476b133
...
+
+    ??? solution
+        This is a sensible approach:
```
adamrtalbot, line 544: Explain what the collate and collect are doing in turn.
```suggestion
        This is a sensible approach. For each item in the channel, the `flatmap` will collate the files into groups of 3 using `.collate(3)`, then collect them into a single tuple using the id as the first value and the collection of 3 files as the second value.
```

docs/advanced/operators.md
```diff
@@ -455,3 +455,357 @@ Launching `./main.nf` [elegant_rutherford] DSL2 - revision: 2c5476b133
...
+Charles
+Jake
+Gina
+Michael
+Amy
+```
+
+### Collecting to a new location
+
+By default, the collected file is written into the work directory, which makes it suitable for input into a downstream process. If the collected file is an output of the workflow instead of an intermediate, it can be written to a directory of your choosing using the `storeDir` argument:
```
adamrtalbot, line 622: Could be useful to have an example of using it in a downstream channel here, since it's not obvious how to do that.

adamrtalbot (approved):
Some minor clarifications. I know this one has a time pressure so I'll approve in absence of anything blocking.

docs/advanced/operators.md
robsyme: Good suggestion. I'll do this live in the workshop and amend the materials afterwards.

christopher-hakkaart (approved):
🤠

PR #335, opened 2024-05-16
Fix 03_hello_gatk.md

SulimanBD:
Change addresses error from not being able to find 'row.id' which leads to an error when running section 6 as the ID is then passed as null.

Error in question is in GATK_JOINTGENOTYPING:
image


mribeirodantas:
Hey @SulimanBD. Can you please try again the current code and see if you're running into an error? I can't reproduce what you've run into.

christopher-hakkaart:
I can't replicate the error either. I'm suspect it was a stochastic error. Closing this for now but will monitor for the same issue.

PR #371, opened 2024-09-19
Add hello-config module

### LINKED ISSUE #370: Module proposal: hello-config
# hello-config

An additional module that would fit fall after hello-nf-test, but could be done at any point after hello-gatk.

## Aims:
- Teach users how to write and edit config files
- Teach users how to handle and set configuration
- Teach users how to configure the running of a Nextflow pipeline

## Proposal:

1. Create a configuration file containing params
2. Modify the `process` scope to change running of a process
3. Use an optional profile to set configuration

## Key targets:

- Understand how you can configure a Nextflow pipeline
- Understand the different scopes of configuration
- Understand how configuration is applied

## To do:

- Decide what configuration is in scope for this tutorial
- Write!

adamrtalbot:
- First draft of hello-config a module for setting Nextflow configuration

Covers setting up configuration of a Nextflow pipeline including software packaging (conda, docker), process directives (cpus, memory), execution (local, slurm) and profiles.

Resolves #370


docs/hello_nextflow/06_hello_config.md
```diff
@@ -0,0 +1,266 @@
...
+conda.enabled = true
+```
+
+_After:_
+
+```groovy title="nextflow.config" linenums="1"
+profiles {
+    'docker' {
+        docker.enabled = true
+        conda.enabled = false
```
maxulysse, line 85: I think the enabled = false are overkill.
We have that in nf-core, but we might be doing too much

docs/hello_nextflow/06_hello_config.md
```diff
@@ -0,0 +1,266 @@
...
+
+```groovy title="nextflow.config" linenums="1"
+docker.enabled = true
+```
+
+_After:_
+
+```groovy title="nextflow.config" linenums="1"
+docker.enabled = false
+conda.enabled = true
```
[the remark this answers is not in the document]
adamrtalbot: I thought it might be nice to be explicit? but I agree about in the profiles.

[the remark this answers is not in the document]
maxulysse: If you do it, then explain why you do it

docs/hello_nextflow/06_hello_config.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+However, if we add this to our config and run the pipeline we will that includes this error:
+
+```console
+Cannot run program "sbatch"
+```
+
+Nextflow has interpreted that we wish to submit to a Slurm cluster, which requires te use of the command `sbatch`. However, because we are not _actually_ using Slurm the `sbatch` command is not installed therefore there is an error.
```
kenibrewer, line 192:
```suggestion
Nextflow has interpreted that we wish to submit to a Slurm cluster, which requires te use of the command `sbatch`. 
However, because our Gitpod instance doesn't have slurm installed (and isn't connected to a cluster)  this throws an error.
```

docs/hello_nextflow/06_hello_config.md
```diff
@@ -0,0 +1,266 @@
...
+
+If we check inside the `.command.run` file created in the work directory, we can see that Nextflow has created a script to submit the job to Slurm.
+
```
kenibrewer, line 195:
```suggestion
!!! reminder

The output of your nextflow console will have the hash of the work subdirectory, which will differ from the paths shown below.
```

I'm not sure if `reminder` is a real option.

docs/hello_nextflow/06_hello_config.md
```diff
@@ -0,0 +1,266 @@
...
+### ---
+### name: 'SAMTOOLS_INDEX (1)'
+### container: 'community.wave.seqera.io/library/samtools:1.20--b5dfbd93de237464'
+### outputs:
+### - 'reads_father.bam'
+### - 'reads_father.bam.bai'
+### ...
+```
+
+If our process had more directives, such as `clusterOptions`, `cpus`, `memory`, `queue`, and `time`, these would also be included in the `.command.run` file and directly passed to the Slurm execution. They would also be translated to the equivalent options for other executors. This is how Nextflow creates the commands required to correctly submit a job to the sbatch cluster via a single configuration change.
```
kenibrewer, line 212:
```suggestion
If our process had more directives, such as `clusterOptions`, `cpus`, `memory`, `queue`, and `time`, these would also be included in the `.command.run` file and directly passed to the Slurm execution. 
They would also be translated to the equivalent options for other executors. 
This is how Nextflow creates the commands required to correctly submit a job to the sbatch cluster via a single configuration change.
```

docs/hello_nextflow/06_hello_config.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+Now run the pipeline using two profiles, `docker` and `local`:
+
+```bash
+nextflow run seqeralabs/nf-hello-gatk -profile docker,local
+```
+
+Note, we have returned back to the original configuration of using Docker containers with a local execution, but we can now use profiles to switch to a different software packaging system (conda) or a different executor (slurm) with a single command-line option.
```
kenibrewer, line 260:
```suggestion
We have returned to the original configuration of using Docker containers with local execution. 
However, now we can use profiles to switch to a different software packaging system (conda) or a different executor (slurm) with a single command-line option.
```

docs/hello_nextflow/06_hello_config.md
```diff
@@ -0,0 +1,266 @@
...
+
+### 4.3 Seqera Platform
+
+Setting up and configuring Nextflow for different executors can be complex and time-consuming. Seqera Platform simplifies this process by providing a user-friendly interface to manage and run Nextflow pipelines on various compute platforms. With Seqera Platform, you can easily configure and launch your pipelines on local machines, HPC clusters, cloud environments, and Kubernetes clusters. The platform abstracts the underlying infrastructure, allowing you to focus on your analysis and research without worrying about the technical details of setting up and managing the execution environment.
```
kenibrewer:
```suggestion
Setting up and configuring Nextflow for different executors can be complex and time-consuming. 
Seqera Platform simplifies this process by providing a user-friendly interface for managing and running Nextflow pipelines on various compute platforms. 
With Seqera Platform, you can easily configure and launch your pipelines on local machines, HPC clusters, cloud environments, and Kubernetes clusters. 
The platform abstracts the underlying infrastructure, allowing you to focus on your analysis and research without worrying about the technical details of setting up and managing the execution environment.
```

kenibrewer (commented):
Really like the flow on this module. Great addition!

docs/hello_nextflow/06_hello_config.md
adamrtalbot:
```suggestion
Nextflow has interpreted that we wish to submit to a Slurm cluster, which requires the use of the command `sbatch`. 
However, because our Gitpod instance doesn't have slurm installed (and isn't connected to a cluster)  this throws an error.
```

docs/hello_nextflow/06_hello_config.md
adamrtalbot: I think it would have to be note.
```suggestion	
!!! note
	
	The output of your nextflow console will have the hash of the work subdirectory, which will differ from the paths shown below.
```

docs/hello_nextflow/06_hello_config.md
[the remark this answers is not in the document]
adamrtalbot: Done with 6d1a8397a867719cc217a57f65b44e1de3ba68cf

docs/hello_nextflow/06_hello_config.md
adamrtalbot: I've removed the Seqera Platform bit so this is no longer relevant.

docs/hello_nextflow/06_hello_config.md
adamrtalbot: I've simplified it to not include those bits!

vdauwera:
Looks great overall, some thoughts for future improvements TBC but this seems great as a first pass to try on a cohort of learners

PR #375, opened 2024-09-23
Fixes for cache_and_resume.md

pinin4fjords:
A couple of fixes:

- Dollar signs are unhelpful in command blocks (they have to be removed by anyone copying/ pasting the blocks)
- Final example is hard to talk through, because it's incomplete- suggesting a fix

In that final example the 'solution' part does not match the 'problem' part, which makes it hard to follow. There's also a lot of 'imagining' required. I've made a simpler, more complete example for both.

### DIFF
--- a/docs/basic_training/cache_and_resume.md
+++ b/docs/basic_training/cache_and_resume.md
@@ -402,7 +402,18 @@ workflow {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


kenibrewer (approved):
Good improvement!

maxulysse (approved):

kenibrewer (approved):

PR #382, opened 2024-10-02
Minor fix to containers

pinin4fjords:
Two minor fixes:

 - First containers example should not contain 'cowsay' since that's the objective of the following exercise.
 - Use literal 'my-image' rather than '<my-image>' all the way through. Right now we do it later on, but not at first, which makes things hard to follow.


maxulysse (approved):

pinin4fjords:
Thanks @maxulysse !

PR #383, opened 2024-10-02
Be more explicit in the conda env creation, be consistent with Seqera Containers/ Biocontainers

pinin4fjords:
- Current conda example a bit obtuse
- Use of Seqera Containers is confusing given we've only mentioned Biocontainers


kenibrewer (approved):
LGTM

PR #385, opened 2024-10-03
Add exercise around the caching example

pinin4fjords:
I tweaked the final part of this session previously, but I think it would benefit from a little exercise to bring the point home.


maxulysse (approved):

pinin4fjords:
Thanks @maxulysse

PR #392, opened 2024-10-10
Add linkify

christopher-hakkaart:
`nf-core pipelines create` fails in Gitpod with error `ModuleNotFoundError: Linkify enabled but not installed.`

Adding linkify-it-py to build to fix.


christopher-hakkaart:
Fixed in another PR

Issue #399, opened 2024-10-18
hello-nf-test got broken by ref files update


vdauwera:
Instructions and canned files are no longer matching what's in the gitpod

PR #400, opened 2024-10-18
Update hello-nf-test instructions and canned test files

vdauwera:
Fixes https://github.com/nextflow-io/training/issues/399


vdauwera:
@kenibrewer The update to the ref files had broken the hello-nf-test training module. This fixes the relevant instructions, paths and test files. If we want to push this out (I assume we do, otherwise nf-test will not work as written) we'll need to have participants restart with a new gitpod. Which is not a big deal but it's a choice.

kenibrewer (approved):
LGTM!

PR #411, opened 2024-10-25
Fix SVG code for new Nextflow and Seqera logos

mribeirodantas:
Fix SVG code for the logos of Nextflow and Seqera at the right sidebar.

Before:
Captura de Tela 2024-10-24 às 21 03 08

After:
Captura de Tela 2024-10-24 às 21 04 07


vdauwera:
Oh good catch. Hey should we/ could we add a link to the forum in that sidebar? 

No need to block this on that if it's not trivial, but if it is, would be good to take care of it.

vdauwera (approved):
LGTM (see my comment about adding link to forum for bonus points, but not a blocker to merging)

mribeirodantas:
Yes, we should! :) I will work on that

PR #415, opened 2024-10-25
devcontainer support with docker-outside-docker

adamrtalbot:
This PR adds support for a devcontainer which allows docker-outside-docker, therefore users can run hello-genomics etc.

The only limitation so far is the paths in the samplesheet don't match (`/workspace/gitpod/`), I've tried to fix it but no luck so far. We could add this right now to provide a backup before we solve the gitpod absolute path problem in the repository.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,10 +1,14 @@
 {
     "name": "nfcore",
-    "image": "nfcore/gitpod:latest",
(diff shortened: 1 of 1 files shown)


kenibrewer (approved):
@adamrtalbot I chatted with GG and Marcel. This is a fairly high-risk change, so we can go ahead and merge this as part of the dev branch for testing purposes, but we plan to continue using gitpod during the Summit next week.

PR #418, opened 2024-10-27
Update Hello nf-test

vdauwera:
Bringing the nf-test materials in sync with the rest of the updated training


vdauwera:
I've checked that only Hello nf-test is affected by my changes and have run through everything twice, pretty sure I caught anything that would be an issue.

PR #420, opened 2024-10-27
New module section hello-compute environments

kenibrewer:
This is a work in progress section attempting to allow students to create a simulated compute environment using Tower Agent and something that can run on Gitpod. 

I initially thought this might work as a "simulated" setup launching a Nextflow job designed to fail but after reading through the content, it just leaves a bad taste. I think we should stick with the more documentation-style training content until we can figure out a compute environment setup that works.


vdauwera:
Would be interesting to revisit this as a potential side quest, though it sounds like there's some work that's needed on the infrastructure side to make this possible.

Issue #422, opened 2024-10-27
Add indents to match destination of code blocks


pinin4fjords:
Example processes components in e.g. hello world have no indent:

```
"""
echo '$greeting' > output.txt
"""
```

This means that when users copy them across to indented process bodies, the indents don't match. Just to keep things super smooth, would be nice if they all had indents matching their destination:

```
    """
    echo '$greeting' > output.txt
    """
```

vdauwera:
Indentation is a popular topic! Made a slightly more widely-scoped issue for code conventions here: https://github.com/nextflow-io/training/issues/517

PR #431, opened 2024-10-30
Fixes Add indents to match destination of code blocks #422

mribeirodantas:
Signed-off-by: Marcel Ribeiro-Dantas <mribeirodantas@seqera.io>


mribeirodantas:
I wonder if we shouldn't stick to nf-core conventions on code indentation, i.e.
```
process FOO {
    publishDir ...
    some other directive ...

    input:
    val foo
    path bar

    output:
    stdout
```
instead of
```
 process FOO {
    publishDir ...
    some other directive ...

   input:
    val foo
    path bar

    output:
    stdout
```

It's not clear to me if the way we're doing contributes a lot for readability/understanding, while we teach people differently from how most of the community writes Nextflow pipelines.

PR #433, opened 2024-11-04
Add an outdir parameter for 'hello genomics' onwards

pinin4fjords:
Training participants during summit highlighted that the `outdir` in the nf-test part had no effect. That's because the workflow didn't have an outdir parameter.

I've removed the hardcoded publishDir back as far as 'hello genomics'.


kenibrewer (approved):
Nice catch!

pinin4fjords:
@kenibrewer I re-requested review, since I extended this all the way back the start of hello genomics

mribeirodantas (dismissed):

kenibrewer (approved):

PR #435, opened 2024-11-04
Add emits to outputs of processes with multiple outputs

pinin4fjords:
Suggest we use `emit:` on the multi outputs so participants done't think they'r stuck with `FOO.out[0]`.


robsyme:
Strongly agree. Any process with more than one output channel should always use `emit`. Better readability for the humans, and much less likely to run into indexing mistakes.

adamrtalbot (dismissed):
I like a lot. Just wait for a nod from one of the advocates.

mribeirodantas (approved):

PR #436, opened 2024-11-05
Update to nf-test

adamrtalbot:
Updates to nf-test which should help clarity

Changes:
 - Move the pipeline test first, to give a gentle introduction
 - Ramp up the complexity throughout the module
 - Add more description of why testing is important

Note I think we could go further, but these are quick and relatively simple changes.


kenibrewer:
@adamrtalbot I like moving the pipeline tests to the beginning. Looks like this needs some updating based on the other changes, though.

vdauwera:
Closing since the rewrite work has started in a separate branch

PR #439, opened 2024-11-06
Apply automatic formatting with the vscode extension

adamrtalbot:


mribeirodantas:
I think it's a bit early to do this. The formatter has been getting some feedback recently and I believe it will have changes in the near future. For indenting, for example, I think we should stick to Harshil's style.

kenibrewer:
Are there any downsides to applying what the formatter recommends right now
and then switching later for the new formatted version?

vdauwera:
Oof we're going to have to redo all of this now that we've merged the new version, aren't we... 

It seems like a lot of it is indentation, which rolls up to https://github.com/nextflow-io/training/issues/517

Going to close this since resolving the merge conflicts would be more work than just doing it from scratch, but I agree we should definitely align on the vscode linting for these conventions.

PR #443, opened 2024-11-12
Add nf-core intermediate training

FriederikeHanssen:
**WIP** (still very raw, no use in detailed reviews just yet, just for the overall structure):
 
  This PR adds nf-core intermediate training. 
  
- Create a new pipeline using a reduced template
- Add a nf-core module
- Create a custom module and add it
 
Naming and placing can still be discussed.

### DIFF
--- a/docs/hello_nextflow/11_hello_nf-core-pipeline.md
+++ b/docs/hello_nextflow/11_hello_nf-core-pipeline.md
@@ -26,7 +26,7 @@ nf-core --help
 
 Here we will focus on the tooling to assist pipeline developers, starting with the `nf-core pipelines create` command.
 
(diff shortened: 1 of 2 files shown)


kenibrewer:
@vdauwera I had imagined that this content would go into our existing nf-core module rather than having a separate intermediate training. Is that what you had in mind too?

vdauwera:
Yes, however I think we need a planning session to think through the scope and content of the current training modules before we continue adding anything major. Let's schedule that asap so we can figure this out from a whole-course perspective.

FriederikeHanssen:
>  I had imagined that this content would go into our existing nf-core module rather than having a separate intermediate training. Is that what you had in mind too?

I don't really mind where to put this. I just needed a place for the moment. I wasn't sure if you wanted to expand the existing nf-core training or not. and what of this will be moved over to nf-core eventually.

FriederikeHanssen:
Moved it over to #448. broke something with git submodules

PR #450, opened 2024-11-15
replace view with dump

FriederikeHanssen:


docs/hello_nextflow/02_hello_world.md
```diff
@@ -1100,35 +1100,41 @@ greeting_ch = Channel.fromPath(params.input_file)
...
+[DUMP: After splitCsv] ['Hello', 'Bonjour', 'Holà']
+[DUMP: After flatten] 'Hello'
+[DUMP: After flatten] 'Bonjour'
+[DUMP: After flatten] 'Holà'
 ```
 
 As you can see, the flatten() operator has transformed the channel from containing arrays to containing individual elements. This can be useful when you want to process each item separately in your workflow.
 
-Remove the `.view()` operations before you continue.
+Remove the `.dump()` operations before you continue.
```
pinin4fjords, line 1137: Maybe we don't need to with dump?

pinin4fjords (approved):
Not 100% sold on the dump thing, I quite often chuck views in as a lightweight thing in development. I have no intention of leaving them there, so dump isn't all that useful, and I have to do something extra on the command line. 

... but I saw on Slack that this is the consensus view, so I'm fine with it.

docs/hello_nextflow/02_hello_world.md
FriederikeHanssen: Rephrased to make it optional. I am terrified of updating all line numbers everywhere.

adamrtalbot (commented):
I'm with @pinin4fjords, not 100% keen on dump replacing view here, it seems more work than necessary?

vdauwera:
After various live discussions we decided to keep `view()` in the beginner training, so I'm closing this, but I'd love to see a side quest on debugging/troubleshooting. @christopher-hakkaart had written a few troubleshooting cases that made for a promising start, which we could repurpose into something with more explicit structure / systematic methodology. 

Made an issue to follow up: https://github.com/nextflow-io/training/issues/519

PR #451, opened 2024-11-15
More specific workspaceMounts for devcontainer

robsyme:
This fix resolves an issue where Gitpod was mounting the repository at `/workspaces/workspaces/training`, which causes issues with the Docker mounting. 

This change in the devcontainers.json mounts the training material at the same location in the container as it exists in the host, which ensures that mount paths to docker when run inside the container also resolve in the host.


pinin4fjords (approved):
Confirmed as working

PR #452, opened 2024-11-16
Fix channel construction for greeting input

ejseqera:


mribeirodantas:
I wonder if the current version isn't a good way to display early that `params` aren't the only way you can feed something to a channel. In the current version, we show it can be done with a simple variable and then, a few lines down the line, we ask them to change this into `params.input_file`. It's not a big change anyway, and I'd be fine merging this but what do you think?

mribeirodantas (approved):

mribeirodantas:
Good catch! 🥳

PR #456, opened 2024-11-17
Correct cowsay -> quote transition

pinin4fjords:
Looks like someone got partway through changing one example to use `quote` instead of `cowsay`. Fixing it up here.

PR #457, opened 2024-11-17
Fix images

pinin4fjords:
Further to https://github.com/nextflow-io/training/pull/456, fix images as @kenibrewer requested


kenibrewer (approved):
Looks great! Thanks!

PR #458, opened 2024-11-17
Executable-fixes

robsyme:
At present, there is a mix of executable and non-executable files in the training materials. 

This has the potential to confuse some workshop participants as the files will be coloured differently in some console outputs (tree, ls, etc).

At no point do we run the scripts as executables (`./main.nf`), so the executable bit is unnecessary.

This PR makes:
- Removes the executable bit from all Nextflow scripts, and
- Removes the executable bit from other non-executable files (fasta, gz, fq, etc)


pinin4fjords (approved):

mribeirodantas:
Good point, @robsyme ! I will merge it, as I don't see any drawback from that. Thanks!

robsyme:
Thanks @mribeirodantas  ❤️

PR #459, opened 2024-11-17
No longer any need to pin to an outdated edge version of Nextflow.

robsyme:
At present, the Dockerfile sets the NXF_VER and NXF_EDGE environment variables, which causes ugly message

```
Nextflow 24.09.2-edge is available - Please consider updating your version to it
```

For all runs. Unpinning these versions (and rebuilding the container) would remove these warnings.

### DIFF
--- a/.github/gitpod.Dockerfile
+++ b/.github/gitpod.Dockerfile
@@ -47,8 +47,8 @@ RUN curl -fSL https://github.com/seqeralabs/tower-agent/releases/latest/download
 USER gitpod
 
 # Uncomment if we need to pin the Nextflow version
-ENV NXF_EDGE=1
-ENV NXF_VER=24.02.0-edge
+# ENV NXF_EDGE=1
+# ENV NXF_VER=24.02.0-edge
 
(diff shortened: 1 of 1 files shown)


pinin4fjords:
We want to pin to _a_ version though, right? Otherwise output could change wrt examples? The older docs even had a bit telling users to do the pinning as I recall, for that reason.

robsyme:
Do you mean output like:
```
 N E X T F L O W   ~  version 24.02.0-edge

 ┃ Launching `hello-genomics.nf` [special_goldstine] DSL2 - revision: 4cbbf6ea3e

executor >  local (6)
[d6/10c2c4] SAMTOOLS_INDEX (1)       | 3 of 3 ✔
[88/1783aa] GATK_HAPLOTYPECALLER (2) | 3 of 3 ✔
```

I would doubt that anybody would be too confused, but I'm happy to comprimise. I'm happy to pin, but would prefer that we pin to a stable release.

robsyme:
Closing in favor of https://github.com/nextflow-io/training/pull/463

PR #461, opened 2024-11-17
Small improvements to add additional context/explanations

ejseqera:
Optional enhancement: Added additional context to the body and notes to provide clearer explanations around concepts such as tuples and the handling of accessory files.


mribeirodantas (commented):
I agree it's an improvement, but the ending worries me a bit. They will be converted into channels implicitly so I think there should be a better way to express the last mention of channels in the new version. What do you think?

ejseqera:
I see your points. I believe that there’s room to provide more context for newer developers about when and why to use the `file()` syntax in scenarios like this. How about reframing it like this:

**"While main data inputs are streamed dynamically through channels, there are two approaches for handling accessory files. The recommended approach is to create explicit channels, which makes data flow clearer and more consistent. Alternatively, the `file()` function to create variables can be used for simpler cases, particularly when you need to reference the same file in multiple processes - though be aware this still creates channels implicitly. For example: ..."**

This way, we emphasize explicit channel creation as the best practice while still introducing `file()` as an alternative for convenience. Do you think this balances the guidance appropriately?

mribeirodantas:
That's an improvement! I keep wanting to throw in the queue/value channel concepts but for very beginners, I think they get more in the way than help 😆. What about a shorter but direct version, like:

```
Main data inputs are streamed through channels, but accessory files that will be used together with multiple different samples can be handled in two ways:

* Recommended: Use explicit channel factories such as `Channel.value` for clear and consistent data flow.
* Alternative: Use the file() function. Note that this creates a channel implicitly.
``

File: docs/hello_nextflow/04_hello_genomics.md [hunk 1]
```diff
@@ -358,7 +358,7 @@ params.intervals        = "${projectDir}/data/ref/intervals.bed"
 
 ### 2.3. Create variables to hold the accessory file paths
 
-Unlike the main data inputs, which must be fed to processes through channels, the accessory files can be handled a bit more simply: we can use the `file()` function to create variables to hold those file paths.
+Accessory files like reference genomes, index files, and interval files are typically static inputs that remain constant throughout execution. Unlike main data inputs, which are streamed dynamically through channels, accessory files can be handled more simply by using the `file()` function to convert file paths into managed file objects without the need for channels.
```
adamrtalbot, line 361:
```suggestion
While main data inputs are streamed dynamically through channels, there are two approaches for handling accessory files. The recommended approach is to create explicit channels, which makes data flow clearer and more consistent. Alternatively, the file() function to create variables can be used for simpler cases, particularly when you need to reference the same file in multiple processes - though be aware this still creates channels implicitly.
```

pinin4fjords:
> In the future, value channels will be deprecated and file

TIL. What about single-value non-file channels?

docs/hello_nextflow/04_hello_genomics.md
pinin4fjords: Like this!

adamrtalbot: I just copy-pasted it from @ejseqera but added it as a suggestion so she can hit commit if she's travelling - don't like adding to other people's PRs.

pinin4fjords (approved):

PR #462, opened 2024-11-18
Adjust paths so that Gitpod (original) mounting points 

robsyme:
Symlink to required so that paths in the materials that expect the Gitpod (classic) paths still resolve in devcontainers.

In the medium term, we should remove paths and that are specific to the original Gitpod implementation.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -34,5 +34,6 @@
             "label": "Application",
             "onAutoForward": "openPreview"
         }
-    }
+    },
+    "postCreateCommand": "mkdir -p /workspace && ln -s /workspaces /workspace/gitpod"
 }


pinin4fjords:
> Does this work with the parts where users [run a container manually](https://training.nextflow.io/hello_nextflow/04_hello_genomics/#022-spin-up-the-gatk-container-interactively)?
> 
> ```
> docker run -it -v ./data:/data community.wave.seqera.io/library/gatk4:4.5.0.0--730ee8817e436867
> ```

Not if it's run from the symlink directory, no, the path is not valid for docker outside docker.

pinin4fjords (approved):
Looks good to me. Makes the bam file with gitpod paths work as an input at least. 

Some examples will be confusing to users due to the mismatch, so we shouldn't leave the proper fix too long.

pinin4fjords:
> Does this work with the parts where users [run a container manually](https://training.nextflow.io/hello_nextflow/04_hello_genomics/#022-spin-up-the-gatk-container-interactively)?
> 
> ```
> docker run -it -v ./data:/data community.wave.seqera.io/library/gatk4:4.5.0.0--730ee8817e436867
> ```

Will also be a problem [here](https://github.com/nextflow-io/training/blob/fa92ba58e71b17f683f6d8256379c1c53f602ad8/docs/hello_nextflow/03_hello_containers.md?plain=1#L118)

adamrtalbot (approved):
neat

PR #463, opened 2024-11-18
Update Nextflow version to latest stable (24.10.0)

robsyme:
I would argue that training should not be done on an edge version of Nextflow. This was originally set so that we could get early access to Phil's very fancy coloured outputs. This feature has been in stable Nextflow for some time, so we can safely move to a stable version of Nextflow.

This will also (temporarily) prevent the ugly message about a new Nextflow version on every run.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,4 @@
+FROM ghcr.io/nextflow-io/training:latest
+
+ENV NXF_VER=24.10.0
+ENV NXF_EDGE=0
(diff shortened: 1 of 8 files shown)


mribeirodantas:
There was some amount of training sessions in the past days/weeks so we were worried of troubles coming up if we had changed to a new version. I'm fine with pining to latest stable release as far as we have time to make sure everything is working before the next session 😅

mribeirodantas:
We have also discussed having a release for every time we update/test the training to the latest stable Nextflow release at the time.

PR #464, opened 2024-11-18
Make cd less specific

pinin4fjords:
This is a fudge, but while we transition to GitPod Flex/ devcontainers, it would be helpful in the first instance if we didn't direct users to do breaking 'cd' operations.

While we have symlinked the alternative gitpod location in https://github.com/nextflow-io/training/pull/462, working within those symlinks will not work (e.g. paths are not valid when we do docker-outside-docker). 

So it would be good if we kept the instructions more generic.

### DIFF
--- a/docs/hello_nextflow/03_hello_containers.md
+++ b/docs/hello_nextflow/03_hello_containers.md
@@ -112,11 +112,7 @@ When you run a container, it is isolated from the host system by default.
 This means that the container can't access any files on the host system unless you explicitly tell it to.
 One way to do this is to **mount** a **volume** from the host system into the container.
 
-Prior to working on the next task, confirm that you are in the `hello-nextflow` directory.
-
-```bash
-cd /workspace/gitpod/hello-nextflow
-```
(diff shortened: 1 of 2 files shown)


docs/hello_nextflow/03_hello_containers.md
```diff
@@ -112,11 +112,7 @@ When you run a container, it is isolated from the host system by default.
 This means that the container can't access any files on the host system unless you explicitly tell it to.
 One way to do this is to **mount** a **volume** from the host system into the container.
 
-Prior to working on the next task, confirm that you are in the `hello-nextflow` directory.
-
-```bash
-cd /workspace/gitpod/hello-nextflow
-```
+Prior to working on the next task, confirm that you are in the `hello-nextflow` directory. Thelast part of the path shown when you type `pwd` should be `hello-nextflow`.
```
FriederikeHanssen, line 115:
```suggestion
Prior to working on the next task, confirm that you are in the `hello-nextflow` directory. The last part of the path shown when you type `pwd` should be `hello-nextflow`.
```

adamrtalbot (approved):

PR #471, opened 2024-11-20
Replace numeric with named outputs for GATK_HAPLOTYPECALLER

adamrtalbot:
Some previous outputs of GATK_HAPLOTYPECALLER were still integer based instead of named. This PR replaces them.


pinin4fjords (approved):
All look good to me. I've checked the process definitions associated with all of these to verify presence of the `emit`s.

Also did a final grep to check there aren't more affected files.

Sorry for not doing these when I did the naming.

PR #472, opened 2024-11-20
Move parameters to above profiles for correct handling of params

adamrtalbot:
Parameters were being read as defaults instead of the profile `demo` which was causing problems. This PR moves them to the right location.


pinin4fjords (approved):
Looks uncontroversial to me

kenibrewer (approved):
Nice catch!

PR #481, opened 2024-11-25
update paths to work with flex

FriederikeHanssen:


pinin4fjords:
https://github.com/nextflow-io/training/pull/462/files was intended to prevent the need for this

pinin4fjords:
... but I do observe the path problems. We do need to maintain the GitPod-compatible links for now, so maybe https://github.com/nextflow-io/training/pull/482 is a better fix (once we double-check).

vdauwera:
This got done as part of the switch to codespaces

PR #482, opened 2024-11-25
Update devcontainer.json for new path structure

pinin4fjords:
Recently created GitPod flex envs suggest we need to update this symlinking, since e.g. /workspace/gitpod/hello-nextflow/data/bam/reads_mother.bam is no longer valid.

I'm not sure why the change occurred, maybe https://github.com/nextflow-io/training/pull/463, we should probably double-check with a fresh org + project in GitPod Flex to be sure. 

**Edit: this seems to be the correct symlinking for quite a fresh org in Flex**

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -38,5 +38,5 @@
             "onAutoForward": "openPreview"
         }
     },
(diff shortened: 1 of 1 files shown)


vdauwera:
Is this still applicable in light of the switch to codespaces? Otherwise, can we close this?

vdauwera:
Closing but feel free to reopen if i'm wrong about this being outdated

PR #486, opened 2024-12-16
New multi-course structure for training materials

vdauwera:
Major refresh of the training materials

- Pared down and streamlined the Hello Nextflow course (v2)
- Spun out the genomics material into a separate mini-course, Nextflow for Genomics
- Added stubs/placeholders for Nextflow Run and Side Quests
- Removed the experimental and/or unfinished content from the training environment
- Updated and streamlined the training index


docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,448 @@
...
+[10/15498d] sayHello [100%] 1 of 1 ✔
+```
+
+This time, Nextflow has created a new directory called `results/`.
+Our `output.txt` file is in this directory.
+If you check the contents it should match the output in the work subdirectory.
+This is how we move results files outside of the working directories conveniently.
+
+It is also possible to set the `publishDir` directive to make a symbolic link to the file instead of actually copying it.
+This is preferable when you're dealing with very large files.
```
pinin4fjords, line 331:
```suggestion
This is preferable when you're dealing with very large files you don't need to retain longer term.
```

I still maintain that the practical applications of the default symliking are few.

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+```console title="Output"
+ N E X T F L O W   ~  version 24.10.0
+
+ ┃ Launching `hello-channels.nf` [lonely_pare] DSL2 - revision: b9f1d96905
+
+executor >  local (3)
+[3d/1fe62c] sayHello (2)       [100%] 3 of 3 ✔
+```
+
+However... This seems to indicate that '3 of 3' calls were made for the process, which is encouraging, but this only give us one subdirectory path. What's going on?
```
pinin4fjords, line 239:
```suggestion
However... This seems to indicate that '3 of 3' calls were made for the process, which is encouraging, but this only shows us a single run of the process, with one subdirectory path (`3d/1fe62c...`). What's going on?
```

The previous one made me think I should be looking at the actual directories momentarily.

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+
+### 2.3. Run the command again with the `-ansi-log false` option
+
+To expand the logging to display one line per process call, add `-ansi-log false` to the command.
+
+```bash
+nextflow run hello-channels.nf -ansi-log false
+```
+
+This time we see all six work subdirectories listed in the output:
```
pinin4fjords, line 251:
```suggestion
This time we see all three process runs and their associated work subdirectories listed in the output:
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+```console title="Output"
+N E X T F L O W  ~  version 24.02.0-edge
+Launching `hello-channels.nf` [big_woese] DSL2 - revision: 53f20aeb70
+[62/d81e63] Submitted process > sayHello (1)
+[19/507af3] Submitted process > sayHello (2)
+[8a/3126e6] Submitted process > sayHello (3)
+```
+
+That's much better; at least for a simple workflow.
+For a complex workflow, or a large number of inputs, having the full list output to the terminal might get a bit overwhelming.
```
pinin4fjords, line 262:
```suggestion
For a complex workflow, or a large number of inputs, having the full list output to the terminal might get a bit overwhelming, so you might not choose to use `-ansi-log false` in those cases.
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+That being said, we have another problem. If you look in the `results` directory, there is only one file: `output.txt`!
+
+```console title="Directory contents"
+results
+└── output.txt
+```
+
+What's up with that? Shouldn't we be expecting a separate file per input greeting, so three files in all?
+Did all three greetings go into a single file?
+You can check the contents of `output.txt`; you will find only one of the three.
```
pinin4fjords, line 273: Add an example cat command?

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+
+As long as the output files stay there, isolated from the other processes, that is okay.
+But when the `publishDir` directive copies each of them to the same `results` directory, whichever got copied there first gets overwritten by the next one, and so on.
+
+### 2.4. Ensure the output file names will be unique
+
+We can continue publishing all the outputs to the same results directory, but we need to ensure they will have unique names.
+Specifically, we need to modify the first process to generate a file name dynamically so that the final file names will be unique.
+
+So how do we make the file names unique? A common way to do that is to use some unique piece of metadata as part of the file name.
```
pinin4fjords, line 286:
```suggestion
So how do we make the file names unique? A common way to do that is to use some unique piece of metadata from the inputs (received from the input channel) as part of the output file name.
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+}
+```
+
+Make sure to replace `output.txt` in both the output definition and in the `script:` command block.
+
+!!! tip
+
+    In the output definition, you MUST use double quotes around the output filename expression (NOT single quotes), otherwise it will fail.
+
+This should produce a unique output file name every time the process is called.
```
pinin4fjords, line 337:
```suggestion
This should produce a unique output file name every time the process is called, so that it can be distinguished from the outputs from other iterations of the same process in the output directory.
```
Just doubling down, not sure if it's helpful.

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+```
+
+Success! Now we can add as many greetings as we like without worrying about output files being overwritten.
+
+!!! note
+
+    In practice, naming files based on the input data itself is almost always impractical.
+    The better way to generate dynamic filenames is to use a samplesheet contain relevant metadata (such as unique sample IDs) and create a data structure called a 'map'.
+    We can then pass the map to the processes, which can be set up to select an appropriate identifier to generate the filenames.
+    You'll learn how to do that later in your Nextflow training.
```
pinin4fjords, line 375:
```suggestion
    In practice, naming files based on the input data itself is almost always impractical.
    The better way to generate dynamic filenames is to pass metatdata to a process along with the input files. We can derive that metadata from a sample sheet as we're reading the input files. 
    You'll learn how to do that later in your Nextflow training.
```

I think the current note is too complex for someone just getting used to the idea of channels.

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+
+### Takeaway
+
+You know how to feed a batch of multiple input elements through a channel.
```
pinin4fjords, line 379:
```suggestion
You know how to feed multiple input elements through a channel.
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+
+### What's next?
+
+Learn how to make the workflow take a file as its source of input values.
+
+---
+
+## 3. Use CLI parameters to supply input values
+
+We want to be able to specify the input from the command line, since that is the piece that will almost always be different in subsequent runs of the workflow.
```
pinin4fjords, line 389:
```suggestion
When we're actually using our workflow, we want to be able to control its inputs from the command line. 
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,868 @@
...
+
+You know how to use CLI parameters to feed inputs to the workflow.
+
+### What's next?
+
+Learn how to manage channel contents using operators.
+
+---
+
+## 4. Supply a batch of multiple values via the `params` system
```
pinin4fjords, line 509: Would this not be a good place to stop, and have a subsequent (possibly quite short) 'hello operators', using `flatten` and `splitCsv` as exemplars?

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
+# Part 3: Hello Plumbing
+
+Most real-world workflows involve more than one step.
+In this training module, you'll learn how to connect processes together in a multi-step workflow.
+
+This will teach you the Nextflow way of achieving the following:
+
+1. Making data flow from one process to the next
+2. Collecting outputs from multiple process calls into a single process call
+3. Passing more than one input through a channel
```
pinin4fjords, line 10: We just did that in the previous module, right?

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
+# Part 3: Hello Plumbing
```
pinin4fjords, line 1:
```suggestion
# Part 3: Hello Workflows
```
Maybe? We haven't really had a workflow so far, just a single step- this marks the point where we actually move beyond a single process into genuine worflows.

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+
+This will teach you the Nextflow way of achieving the following:
+
+1. Making data flow from one process to the next
+2. Collecting outputs from multiple process calls into a single process call
+3. Passing more than one input through a channel
+4. Handling multiple outputs coming out of a process
+
+To demonstrate, we will continue building on the domain-agnostic Hello World example from Parts 1 and 2.
+This time, we're going to make the following changes to our workflow:
```
pinin4fjords, line 14:
```suggestion
This time, we're going to make the following changes to our workflow to better reflect how people build actual workflows:
```

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+
+1. Add a second step that converts the greeting to uppercase;
+2. Add a third step that collects all the transformed greetings and writes them into a single file;
+3. Add a parameter to name the final output file and pass that as a secondary input to the collection step;
```
pinin4fjords, line 18:
```suggestion
1. Add a second step that converts the greeting to uppercase.
2. Add a third step that collects all the transformed greetings and writes them into a single file.
3. Add a parameter to name the final output file and pass that as a secondary input to the collection step.
```

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+
+executor >  local (3)
+[74/e135b2] sayHello (3)       [100%] 3 of 3 ✔
+```
+
+---
+
+## 1. Add a second step to the workflow
+
+We want to add a second step to convert the greeting to uppercase, using the classic UNIX text replacement command `tr '[a-z]' '[A-Z]'`.
```
pinin4fjords, line 49:
```suggestion
Real workflows have more than one process! So it's time to start adding some more.

We'll add a second step to convert the greeting to uppercase, using the classic UNIX text replacement command `tr '[a-z]' '[A-Z]'`.
```

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+
+First, we need to write a new process that wraps the `tr '[a-z]' '[A-Z]'` command.
+Then we'll need to add it to the workflow, setting it up to take the output of the `sayHello()` process as input.
+
+### 1.0. Run the uppercasing command in the terminal
+
+The step we want to add to our workflow will use the text replacement command `tr '[a-z]' '[A-Z]'` to convert the greetings output by the first step to uppercase.
```
pinin4fjords, line 56:
```suggestion
The step we want to add to our workflow will use the text replacement command `tr '[a-z]' '[A-Z]'` to convert the greetings output by the first step to uppercase.
```
Repetitive with above.

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+[45/eb4757] sayHello (2)       [100%] 3 of 3, cached: 3 ✔
+[ae/4579ab] convertToUpper (3) [100%] 3 of 3 ✔
+```
+
+Have a look inside the work directory of one of the calls to the second process.
+
+```bash
+tree -a work/ae/4579ab5b4f2c1d986d3a955e31f2b7/
+```
+
```
pinin4fjords, line 189:
```suggestion
(you'll need to adapt this tree command to the actual directory name you see in the output from your workflow run)

```

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+```console title="Output"
+work/ae/4579ab5b4f2c1d986d3a955e31f2b7/
+├── Holà-output.txt -> /workspace/gitpod/hello-nextflow/work/dc/93eab52bd47ef198b1cfe1a7721b4b/Holà-output.txt
+└── UPPER-Holà-output.txt
+```
+
+The output of the first process is in there because Nextflow staged it there in order to have everything needed for execution within the same subdirectory.
+However, it is actually a symbolic link pointing to the the original file in the subdirectory of the first process call.
+By default, Nextflow uses symbolic links rather than copies to stage input and intermediate files.
```
pinin4fjords, line 200:
```suggestion
By default, when running on a single machine as we're doing here, Nextflow uses symbolic links rather than copies to stage input and intermediate files.
```

Gets more complex when we start talking about cloud etc.

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+
+### What's next?
+
+Learn how to collect outputs from batched process calls and feed them into a single process.
+
+---
+
+## 2. Add a third step to collect all the greetings
+
+When we apply a transformation to a batch of inputs, like we're doing here to the multiple greetings, we'll often want to collect the transformed outputs and feed them into a single step that performs some kind of analysis or summation.
```
pinin4fjords, line 222:
```suggestion
When we use a process to apply a transformation to each of the elements in a channel, like we're doing here to the multiple greetings, we sometimes want to collect elements from the output channel of that process, and feed them into another process that performs some kind of analysis or summation.
```

I think it's worth being specific with the channel/ process language, since we've introduced those concepts, so that people get used to them.

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+
+Here we're simply going to write them all out to a single file, using the UNIX `cat` command.
```
pinin4fjords, line 224:
```suggestion
In the next step we're simply going to write all the elements of a channel to a single file, using the UNIX `cat` command.
```

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+        path "COLLECTED-output.txt"
+
+    script:
+    """
+    ??? > 'COLLECTED-output.txt'
+    """
+}
+```
+
+This is what we can write based on what you've learned so far.
```
pinin4fjords, line 278: Is this really true? We already know about variable path inputs, and the syntax isn't different for multiple files.

docs/hello_nextflow/03_hello_plumbing.md
```diff
@@ -0,0 +1,755 @@
...
+
+```console title="Output"
+HELLO
+BONJOUR
+HOLà
+```
+
+### Takeaway
+
+You know how to pass more than one input through a channel.
```
pinin4fjords, line 617:
```suggestion
You know how to pass collected inputs through a channel.
```

To dissambuguate from the earlier examples with multiple items.

docs/hello_nextflow/02_hello_channels.md
vdauwera: completely agree

docs/hello_nextflow/03_hello_plumbing.md
vdauwera: Do you mean "Passing more than one input through a channel"? Here it's about having multiple 'tracks' in the input channel, eg the data but also some parameters etc., as a opposed to 'several pieces of the same kind of data'. In general this is a point of confusion for which we should try to develop a controlled vocabulary, because 'multiple inputs' ends up being ambiguous

docs/hello_nextflow/03_hello_plumbing.md
vdauwera: I'm totally open to changing the name. FWIW 'Plumbing' speaks to me because it feels directly evocative of what this one is about, ie specifically how do you connect the pipes. I worry that 'Workflows' is maybe a bit too generic; I see what you're going for but is that going to be obvious to a naive user? They might go "isn't it all about workflows?" Perhaps we can poll for opinions/suggestions.

docs/hello_nextflow/03_hello_plumbing.md
vdauwera: Well we don't know yet that the syntax isn't different. As a naive user, I'm like ok, I'm confident in all of this, but for this next bit I need to check if the syntax is the same or if i need to do something special.

docs/hello_nextflow/03_hello_plumbing.md
vdauwera: I find this _more_ ambiguous though; this reads like it's about the collected greetings, but it's not; it's about that parameter that constitutes an extra 'track' in the input channel. This is exactly what I was talking about in my other comment just now; the ways we talk about multiple inputs is inconsistent and confusing.

docs/hello_nextflow/03_hello_plumbing.md
pinin4fjords: Yes, I actually realised what you meant further down. This probably needs to be something like 'passing collections of elements to a single task' or something similar.

docs/hello_nextflow/03_hello_plumbing.md
pinin4fjords: This was kind of conscious (kind of). I was thinking the first sentence could be along the lines "You thought you'd been thinking about workflows already, but real workflows do these things..."

docs/hello_nextflow/03_hello_plumbing.md
pinin4fjords: To me 'passing more than one input through a channel' covers both things- you're doing it in both cases. The difference here is that we're passing multiple inputs to a single task (a task being a single iteration of a process).

vdauwera: But that's not what this specific line is about; that's not the takeaway. The takeaway from this specific section is that we now know that if the process expects `path foo` and `val bar`, i have to write the process call as `myProcess(foo, bar)`. Before this, all I knew to do was `myProcess(foo)`.

docs/hello_nextflow/02_hello_channels.md
vdauwera: As discussed in slack, I'll add some more explicit text calling out the importance of operators in section 4 and won't split anything out for now.

docs/hello_nextflow/03_hello_plumbing.md
vdauwera: Are you using 'collections of elements' in the same way I'm using 'input tracks in a channel'?

docs/hello_nextflow/03_hello_plumbing.md
pinin4fjords: OK, apologies for the quick comment, I conflated issues while multitasking and I've re-read. 

In fact, the point of this section (section 3) is that we've added a second input to the process, and filled that from a new parameter, right? So both of our versions are wrong. It's nothing to do with multiple inputs through 'a' channel (in fact there are now 2 channels, though one is implicit).

Maybe something like:

```
You know how to add more than one input channel to a process, and use both inputs from the workflow
```

vdauwera: Updated this in a previous pass, thanks for clarifying

docs/hello_nextflow/03_hello_plumbing.md
vdauwera: This was changed to Hello Workflow based on group consensus

docs/hello_nextflow/03_hello_plumbing.md
vdauwera: fixed this in a previous pass

vdauwera:
First green check in two months, thank you @kenibrewer for fixing the pre-commit stuff 🥹

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+    sayHello()
+}
+```
+
+As you can see, a Nextflow script involves two main types of core components: one or more **processes**, and the **workflow** itself.
+Each **process** describes what operation(s) the corresponding step in the pipeline should accomplish, while the **workflow** describes the dataflow logic that connects the various steps.
+
+Let's take a closer look at the **process** block first, then we'll look at the **workflow** block.
+
+### 1.2 The `process` definition
```
mribeirodantas, line 113:
```suggestion
### 1.2 The `process` block
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+This a very minimal process definition that just contains an `output` definition and the `script` to execute.
+
+The `output` definition includes the `path` qualifier, which tells Nextflow this should be handled as a path (includes both directory paths and files).
+Another common qualifier is `val`.
+
+!!! note
+
+    The output definition does not _determine_ what output will be created.
+    It simply _declares_ what is the expected output, so that Nextflow can look for it once execution is complete.
+    This is necessary for verifying that the command was executed successfully and for passing the output to downstream processes if needed.
```
mribeirodantas, line 146:
```suggestion
    This is necessary for verifying that the command was executed successfully and for passing the output to downstream processes if needed. Output produced that doesn't match what is declared in the output block will not be passed to downstream processes.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+!!! warning
+
+    This example is brittle because we hardcoded the output filename in two separate places (the script and the output blocks).
+    If we change one but not the other, the script will break.
+    Later, you'll learn how to use variables to avoid this problem.
+
+In a real-world pipeline, a process usually contains additional blocks such as directives, inputs, and conditional clauses, which we'll introduce later in this training course.
```
mribeirodantas, line 154: Will we? I did a quick search and couldn't find examples of conditional clauses in this training course.

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+```
+
+These are the helper and log files:
+
+- **`.command.begin`**: Metadata related to the beginning of the execution of the process call
+- **`.command.err`**: Error messages (`stderr`) emitted by the process call
+- **`.command.log`**: Complete log output emitted by the process call
+- **`.command.out`**: Regular output (`stdout`) by the process call
+- **`.command.run`**: Full script run by Nextflow to execute the process call
+- **`.command.sh`**: The command that was run by the process call call
```
mribeirodantas, line 266:
```suggestion
- **`.command.sh`**: The command that was actually run by the process call
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+Learn how to manage your workflow executions conveniently.
+
+---
+
+## 3. Manage workflow executions
+
+Knowing how to launch workflows and retrieve outputs is great, but you'll quickly find there are a few other aspects of workflow management that will make your life easier, especially if you're developing your own workflows.
+
+Here we show you how to use the `publishDir` directive to TODO, the `resume` feature for when you need to re-launch the same workflow, and how to delete older work directories with `nextflow clean`.
```
mribeirodantas, line 294:
```suggestion
Here we show you how to use the `publishDir` directive to store in an output folder all the main results from your pipeline run, the `resume` feature for when you need to re-launch the same workflow, and how to delete older work directories with `nextflow clean`.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+### 3.1. Publish outputs
+
+As you have just learned, the output produced by our pipeline is buried in a working directory several layers deep.
+This is done on purpose; Nextflow is in control of this directory and we are not supposed to interact with it.
+
+However, that makes it inconvenient to retrieve outputs that we care about.
+
+Fortunately, Nextflow provides a way to manage this more conveniently, called the `publishDir` directive, which acts at the process level.
+This directive tells Nextflow to copy the output(s) of the process to a designated output directory.
```
mribeirodantas, line 304:
```suggestion
This directive tells Nextflow to publish the output(s) of the process to a designated output directory. By default, the outputs are published as symbolic links from the `work` directory.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+Launching `hello-world.nf` [jovial_mayer] DSL2 - revision: 35bd3425e5
+
+executor >  local (1)
+[62/49a1f8] sayHello | 1 of 1 ✔
+```
+
+This time, Nextflow has created a new directory called `results/`.
+Our `output.txt` file is in this directory.
+If you check the contents it should match the output in the work subdirectory.
+This is how we move results files outside of the working directories conveniently.
```
mribeirodantas, line 353:
```suggestion
This is how we publish results files outside of the working directories conveniently.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+It is also possible to set the `publishDir` directive to make a symbolic link to the file instead of actually copying it.
```
mribeirodantas, line 355:
```suggestion
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+This is preferable when you're dealing with very large files you don't need to retain longer term.
```
mribeirodantas, line 356:
```suggestion
When you're dealing with very large files that you don't need to retain for long, you may prefer to set the `publishDir` directive to make a symbolic link to the file instead of copying it.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+However, if you delete the work directory as part of a cleanup operation, you will lost access to the file, so always make sure you have actual copies of everything you care about before deleting anything.
```
mribeirodantas, line 357:
```suggestion
However, if you delete the work directory as part of a cleanup operation, you will lose access to the file, so always make sure you have actual copies of everything you care about before deleting anything.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+!!! note
+
+    A newer syntax option had been proposed to make it possible to declare and publish workflow-level outputs, documented [here](https://www.nextflow.io/docs/latest/workflow.html#publishing-outputs).
```
mribeirodantas, line 361:
```suggestion
    A newer syntax option documented [here](https://www.nextflow.io/docs/latest/workflow.html#publishing-outputs) has been proposed to make it possible to declare and publish workflow-level outputs.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+```console title="Output" linenums="1"
+ N E X T F L O W   ~  version 24.10.0
+
+Launching `hello-world.nf` [golden_cantor] DSL2 - revision: 35bd3425e5
+
+[62/49a1f8] sayHello | 1 of 1, cached: 1 ✔
+```
+
+Look for the `cached:` bit that has been added in the process status line (line 6), which means that Nextflow has recognized that it has already done this work and simply re-used the result from the previous successful run.
```
mribeirodantas, line 394:
```suggestion
Look for the `cached:` bit that has been added in the process status line (line 5), which means that Nextflow has recognized that it has already done this work and simply re-used the result from the previous successful run.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+!!! note
+
+    When your re-run a pipeline with `resume`, Nextflow does not overwrite any files written to a `publishDir` directory by any process call that was previously run successfully.
+
+### 3.3. Delete older work directories
+
+During the development process, you'll typically run your draft pipelines a large number of times, which can lead to an accumulation of very many files across many subdirectories.
+Since the subdirectories are named randomly, it is difficult to tell from their names what are older vs. more recent runs.
+
+Nextflow includes a convenient `-clean` command that can automatically delete the work subdirectories for past runs that you no longer care about, with several [options](https://www.nextflow.io/docs/latest/reference/cli.html#clean) to control what will be deleted.
```
mribeirodantas, line 408:
```suggestion
Nextflow includes a convenient `clean` subcommand that can automatically delete the work subdirectories for past runs that you no longer care about, with several [options](https://www.nextflow.io/docs/latest/reference/cli.html#clean) to control what will be deleted.
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+Launching `hello-world.nf` [elated_lavoisier] DSL2 - revision: 7c031b42ea
+
+executor >  local (1)
+[4b/654319] sayHello | 1 of 1 ✔
+```
+
+Be sure to open up the output file to check that you now have the new version of the greeting.
+
+```console title="output.txt" linenums="1"
```
mribeirodantas, line 586:
```suggestion
```console title="results/output.txt" linenums="1"
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+Launching `hello-world.nf` [determined_edison] DSL2 - revision: 3539118582
+
+executor >  local (1)
+[72/394147] sayHello | 1 of 1 ✔
+```
+
+Check the output in the results directory:
+
+```console title="output.txt" linenums="1"
```
mribeirodantas, line 639:
```suggestion
```console title="results/output.txt" linenums="1"
```

docs/hello_nextflow/01_hello_world.md
```diff
@@ -0,0 +1,686 @@
...
+
+Launching `hello-world.nf` [elegant_faraday] DSL2 - revision: 3539118582
+
+executor >  local (1)
+[6f/a12a91] sayHello | 1 of 1 ✔
+```
+
+Now you will have the corresponding new output in your results directory.
+
+```console title="output.txt" linenums="1"
```
mribeirodantas, line 668:
```suggestion
```console title="results/output.txt" linenums="1"
```

mribeirodantas (changes requested):
For now, it's a review only of Part I of the Hello Nextflow. ASAP I will review the other parts.

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+
+Launching `hello-channels.nf` [insane_lichterman] DSL2 - revision: c33d41f479
+
+executor >  local (1)
+[86/9efa08] sayHello | 1 of 1 ✔
+```
+
+As previously, you will find the output file named `output.txt` in the `results` directory (specified by the `publishDir` directive).
+
+```console title="output.txt" linenums="1"
```
mribeirodantas, line 40:
```suggestion
```console title="results/output.txt" linenums="1"
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+
+## 1. Provide variable inputs via a channel explicitly
+
+We are going to create a **channel** to pass the variable input to the `sayHello()` process instead of relying on the implicit handling, which has certain limitations.
+
+### 1.1. Create an input channel
+
+There are a variety of **channel factories** that we can use to set up a channel.
+To keep things simple for now, we are going to use the most basic channel factory, called `Channel.of`, which will create a channel containing a single value.
+Functionally this will be exactly equivalent to how we had it set up before, but explicit instead of implicit.
```
mribeirodantas, line 56: `Channel.value(` would be functionally/exactly equivalent to how we had it set up before, but explicit instead of implicit. The implicit conversion to channel creates a value channel. `Channel.of`, which we're using now, creates a queue channel.

I'd rephrase this phrase as something like "Instead of having Nextflow create a channel implicitly, we are doing this explicitly now."

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+
+This is the line of code we're going to use:
+
+```console title="Syntax"
+greeting_ch = Channel.of('Hello Channels!')
+```
+
+This creates a channel called `greeting_ch` using the `Channel.of()` channel factory, which sets up a simple value channel, and loads the string `'Hello Channels!'` to use as the greeting value.
```
mribeirodantas, line 64:
```suggestion
This creates a channel called `greeting_ch` using the `Channel.of()` channel factory, which sets up a simple queue channel, and loads the string `'Hello Channels!'` to use as the greeting value.
```

If you want a value channel with `Channel.of`, you must apply a channel operator that returns a single channel element, such as `last,` `first,` `collect,` etc, e.g `Channel.of('Hello').first()`.

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+
+Launching `hello-channels.nf` [nice_heisenberg] DSL2 - revision: 41b4aeb7e9
+
+executor >  local (1)
+[3b/f2b109] sayHello (1) | 1 of 1 ✔
+```
+
+You can check the results directory to satisfy yourself that the outcome is still the same as previously.
+
+```console title="output.txt" linenums="1"
```
mribeirodantas, line 154:
```suggestion
```console title="results/output.txt" linenums="1"
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+---
+
+## 2. Modify the workflow to run on multiple input values
+
+Workflows typically run on batches of inputs that are meant to be processed in bulk, so we want to upgrade the workflow to accept multiple input values.
+
+### 2.1. Load multiple greetings into the input channel
+
+Conveniently, the `Channel.of()` channel factory we've been using is quite happy to accept more than one value, so we don't need to modify that at all.
+We just have to load more values into the channel.
```
mribeirodantas, line 181: I have always referred to the individual entries in a channel as elements, so as to not make people confused with what we mean by values when we talk about value channels (singletons). As we mention value channels in this part II, I'd rather call these elements. See the example below.

```
Channel.of('A', 'B', '[C, D]')
```
A, B, and [C, D] are three elements in the channel. C and D are items in the third channel.

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+
+executor >  local (3)
+[8e/bb08f3] sayHello (2) | 3 of 3 ✔
+Before flatten: [Hello, Bonjour, Holà]
+After flatten: Hello
+After flatten: Bonjour
+After flatten: Holà
+```
+
+You see that we get a single `Before flatten:` statement because at that point the channel contains one item, the original array.
```
mribeirodantas, line 573: So far, I have seen the words value, element, and item referring to the same thing (what is known as a channel element) in part II. I'd rather stick to elements (and items being the individual entries within an element).

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+    // create a channel for inputs
+    greeting_ch = Channel.of(greetings_array)
+                         .flatten()
+    ```
+
+    We left them in the `hello-channels-3.nf` solution file for reference purposes.
+
+### Takeaway
+
+You know how to use an operator like `flatten()` to transform the contents of a channel, and how to use the `view()` directive to inspect channel contents before and after applying an operator.
```
mribeirodantas, line 592:
```suggestion
You know how to use an operator like `flatten()` to transform the contents of a channel, and how to use the `view()` operator to inspect channel contents before and after applying an operator.
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+```groovy title="hello-channels.nf" linenums="25"
+/*
+ * Pipeline parameters
+ */
+params.greeting = 'greetings.csv'
+```
+
+#### 4.1.2. Switch to a channel factory designed to handle a file
+
+Since we now want to use a file instead of simple values as the input, we can't use the `Channel.of()` channel factory from before.
```
mribeirodantas, line 647:
```suggestion
Since we now want to use a file instead of simple strings as the input, we can't use the `Channel.of()` channel factory from before.
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+The `map()` operator is a very handy little tool that allows us to do all kinds of mappings to the contents of a channel.
+
+In this case, we're going to use it to extract that one element that we want from each line of our file.
+This is what the syntax looks like:
+
+```groovy title="Syntax"
+.map { item -> item[0] }
+```
+
+This means 'for each item in the channel, take the first of any elements it contains'.
```
mribeirodantas, line 788:
```suggestion
This means 'for each element in the channel, take the first of any items it contains'.
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+After map: Hello
+After map: Bonjour
+After map: Holà
+```
+
+Looking at the output of the `view()` statements, we see the following:
+
+- A single `Before splitCsv:` statement: at that point the channel contains one item, the original file path.
+- Three separate `After splitCsv:` statements: one for each greeting, but each is contained within an array that corresponds to that line in the file.
+- Three separate `After map:` statements: one for each greeting, which are now individual items in the channel.
```
mribeirodantas, line 850:
```suggestion
- Three separate `After map:` statements: one for each greeting, which are now individual elements in the channel.
```

docs/hello_nextflow/02_hello_channels.md
```diff
@@ -0,0 +1,880 @@
...
+### 4.1. Modify the script to expect a CSV file as the source of greetings
+
+To get started, we're going to need to make two key changes to the script:
+
+- Switch the input parameter to point to the CSV file
+- Switch to a channel factory designed to handle a file
+
+#### 4.1.1. Switch the input parameter to point to the CSV file
+
+Remember the `params.greeting` parameter we set up in Part1?
```
mribeirodantas, line 622: Sometimes you add a space after Part. Some other times you don't. I'm for consistency + space after Part.
```suggestion
Remember the `params.greeting` parameter we set up in Part 1?
```

mribeirodantas (changes requested):
Part 2 is very nice! I requested some minor changes. Tomorrow I will review the next parts.

docs/hello_nextflow/02_hello_channels.md
vdauwera: Ok I agree with the push for consistency. But it's okay if some instances are left for later; getting the reorg merged soon is more important than having every word be perfect.

docs/hello_nextflow/02_hello_channels.md
vdauwera: ok feel free to suggest changes directly; otherwise let's do this in a refinement pass after this is merged in.

docs/hello_nextflow/02_hello_channels.md
vdauwera: fair point, addressed in commit

mribeirodantas (approved):

maxulysse:
I have just minor comments (but not related to the content itself, as I believe it's just comments related to the content that was ported)

- Why mixing multi lines comments (with the `/*...*/` and single line comments `//`?
  - Nowadays when I comment I use the shortcut from VS code and even multilines are using single comments, much less confusing I believe
- Why not using the tag directive for the processes when we use the flatten operator, so that we see a bit more than (1) (2) or (3)
- Why not using a params.yml over a params.json as json are really not meant to be edited by human hands

PR #488, opened 2025-01-02
🔧 config: add issue template config and bug report template

kenibrewer:
This adds an issue template chooser config (see [docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser) along with a bug report issue template. 

Together these will encourage folks learning Nextflow to get help in the correct location and ensure we have the info needed for looking into real bugs. 

Might be worth adding a new content and/or improvement issue template as part of a separate PR. In the meantime, this config is set to allow blank issues that don't aren't one of the defined issue types (bug-report).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,48 @@
+name: Bug Report
+description: Report a problem in the training materials.
+title: "[Bug]: "
+labels: ["bug", "triage"]
+body:
+    - type: markdown
+      attributes:
+          value: |
+              Thanks for taking the time to report a problem in the training materials.
+              Note: If you are learning Nextflow and need help with the content, ask your questions in the [Seqera community forum](https://community.seqera.io/c/help/37).
+    - type: input
+      id: version
+      attributes:
+          label: Training materials version
+          description: What's the git commit hash of the training materials you are using?
+          placeholder: ex. email@example.com
+      validations:
+          required: false
+    - type: textarea
+      id: whats-wrong
+      attributes:
+          label: What's wrong?
+          description: What's the problem in the training materials?
+          placeholder: Tell us what you see!
+          value: "A bug happened!"
+      validations:
+          required: true
+    - type: textarea
+      id: nextflow-code
+      attributes:
+          label: Nextflow code
+          description: |
+              If relevant, please copy and paste the Nextflow code that you used or constructed following the training materials directions.
+              This will be automatically formatted as Nextflow, so no need for backticks.
+          render: nextflow
+    - type: textarea
+      id: logs
+      attributes:
+          label: Relevant log output
+          description: |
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,48 @@
...
+      attributes:
+          value: |
+              Thanks for taking the time to report a problem in the training materials.
+              Note: If you are learning Nextflow and need help with the content, ask your questions in the [Seqera community forum](https://community.seqera.io/c/help/37).
+    - type: input
+      id: version
+      attributes:
+          label: Training materials version
+          description: What's the git commit hash of the training materials you are using?
+          placeholder: ex. email@example.com
```
vdauwera, line 16: this seems like the wrong kind of placeholder?

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,48 @@
...
+      validations:
+          required: false
+    - type: textarea
+      id: whats-wrong
+      attributes:
+          label: What's wrong?
+          description: What's the problem in the training materials?
```
vdauwera, line 23: This seems a bit redundant with the label, maybe state the purpose, ie are we looking for a high level summary here?

vdauwera (commented):
I love the idea of an issue template. I'm wondering however  if this is maybe a little over-engineered, ie separating problem description / code / log output into separate fields might be a bit too rigid. If someone tried several different forms of the code, how/where would they put that? I would tend to favor having one big area for people to describe the problem, and maybe have the placeholder suggest the content/structure we need in order to troubleshoot (ie instructions to include any relevant code and log output). 

Also feels like this is missing requesting the specific page URL with the instructions they were following when things went wrong.

mribeirodantas (approved):

vdauwera:
Revisiting this and wondering if we would be better off pointing people to the forum for all questions, and opening issues ourselves if we determine there's something actually wrong in the code. 

Thoughts?

ewels:
Old PR, but I think I'm sort of halfway inbetween:

* Point to the community forum (this PR already introduces this, it has a link in the issues config)
* Keep the issue template simpler, or even just blank, just try to make it clear that it's for reporting bugs and not getting help. Not sure that we really need all the different fields.

vdauwera:
Honestly we might just not need a template. We haven't yet received an issue where I've thought "gee you know what I wish this person had been provided a template."

vdauwera:
I wouldn't want us to have one just to have one and then it just gets in the way, you know?

ewels:
Yeah that's kind of what I was thinking 👍🏻  Then we just want the thing to show a link to the forum then, and allow a blank issue.

vdauwera:
Yeah that sounds good

PR #490, opened 2025-01-14
Translate the Environment setup section to Brazilian portuguese

jvfe:
Per the title!

Also translated the orientation section from the Fundamentals training.

Having the fundamentals training fully in portuguese is already great, but the env setup makes it seem more complete.

### DIFF
--- a/docs/basic_training/orientation.pt.md
+++ b/docs/basic_training/orientation.pt.md
@@ -1,0 +1,93 @@
+---
+title: Orientação
+description: Como configurar um ambiente de desenvolvimento para executar Nextflow
+---
+
+# Orientação
+
+O ambiente do Gitpod contem dados de teste que serão utilizados nesse treinamento.
+
+!!! note
+
+    Vá [para este link](../envsetup/index.md) se você ainda não configurou seu ambiente no Gitpod.
+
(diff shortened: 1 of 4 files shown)


docs/basic_training/orientation.pt.md
```diff
@@ -0,0 +1,93 @@
...
+!!! question "Exercise"
+
+    Abra o [ambiente de treinamento no Gitpod](https://gitpod.io/#https://github.com/nextflow-io/training) e use o seguinte comando para ir até a pasta `nf-customize`. Visualize os arquivos nessa pasta utilizando o comando `tree`:
+
+    ```bash
+    cd /workspace/gitpod/nf-training
+    tree .
+    ```
+
+## Variáveis do ambiente
```
jvfe, line 75: I did find it weird that this section is almost word-for-word the same as the section above in the original document, but there's nothing I can do translation-wise about this.

mribeirodantas: Good point 🤯. It's been a while we focused on the Hello Nextflow series and stopped working on this one but that should be fixed, for sure. I will open an issue.

mribeirodantas (approved):

PR #500, opened 2025-02-13
Point `workspaceFolder` to `/workspaces/training/hello-nextflow/`

maxulysse:
Continuing a discussion with @robsyme 
We played with GitHub Codespaces and we noticed that it was errorring and not expending the value from `${localWorkspaceFolder}`.

I propose to point directly to `/workspaces/training/hello-nextflow/`, so that we open the training with the hello-nextflow folder, not to confuse people with tons of folder and files.

Screenshot from 2025-02-13 16-04-18

People doing other part of the training should know enough to change workspaces (and we should document it)


maxulysse:
> The only issue here is that this hard-codes the hello-nextflow series as the only workspace, and wouldn't work for other training materials. I think we need to find a way to specify the workspace.

I think we need to see it in a different way.
We have a quick intro to bash and CLI, let's add a quick intro to VSCode and explain how to change workspace, but by default you're in this small one

pinin4fjords:
It's honestly not that hard to jump into a folder and exclude the rest, @adamrtalbot had people doing that in the last training I was on with him, even for folders within 'hello'. In the absence of a linkable way of specifying a workspace, it's better people start at the top and we show them how to do that (IMO).

.devcontainer/devcontainer.json
```diff
@@ -8,7 +8,7 @@
     "features": {
         "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {}
     },
-    "workspaceFolder": "${localWorkspaceFolder}",
+    "workspaceFolder": "/workspaces/training/hello-nextflow/",
```
maxulysse, line 11:
```suggestion
    "workspaceFolder": "/workspaces/training/",
```

pinin4fjords:
image

vdauwera:
> Merging now and asking forgiveness later 🤠

Doh, I missed this. Two things:

1. Please check with all stakeholders before changing a default, even if it's reverting a behavior to how it had been previously at some point. 

2. I'm pleased to see this isn't actually working as described, because in the meantime some of us discussed this and (in line with @pinin4fjords 's comment above) we agreed to keep the entrypoint to be the top level directory and just document that people have to `cd` into the correct directory. 

Corollary: either this got reverted or what was put in doesn't work? I'd like to know which it is, to make sure we understand current behavior and nothing is going to changed from under our feet.

ewels:
1. This was a hard failure before this PR - GitHub Codespaces wouldn't load at all. It threw a big ugly error and then rebooted with the default settings and container. Hence it needed an urgent merge in order to unblock usage of Codespaces.
2. The PR title is wrong. Maxime did check on slack and changed the PR to go to the root folder instead of nf-hello. So no change in behaviour. So it is working as intended. 

So it didn't get reverted and it didn't not work - it was changed to not follow the initial title. If you check the files diff on the PR you'll see the file path doesn't include nf-hello.

PR #501, opened 2025-02-14
💚 ci(pre-commit): replace prettier workflow with pre-commit

kenibrewer:
This replaces the previous prettier github workflow with a pre-commit run which will ensure consistency between local pre-commit runs and our ci. In addition, this removes the .pretterrc.yml file and replaces it with an editorconfig file that matches the config used in nf-core repos. Finally, this applies the new editorconfig settings to the repo as a whole.


vdauwera (approved):
LGTM

PR #502, opened 2025-02-14
No nav tabs

ewels:
Stacked on top of https://github.com/nextflow-io/training/pull/486

One-line change to remove `- navigation.tabs` from `mkdocs.yml` so that we can see how the nav would look without the Navigation Tabs top-nav.


vdauwera (approved):
LGTM

PR #505, opened 2025-02-17
Update 01_orientation.md

maxulysse:


maxulysse:
This is changed in @vdauwera new version.
But wrong in current version

vdauwera (approved):
good catch

PR #511, opened 2025-02-18
Minor fixes

ewels:
A few tweaks to catch minor things spotted on first impressions with the newly updated docs.


ewels:
⚠️ Needs testing. Tried just now and got an error that the workspace folder didn't exist. Not sure if a one-off of there's something in these files that was being used.

ewels:
Merging into a branch as I think testing issues may be due to me running on a fork.

ewels:
Yes, was a fork / repo name problem, works fine still when coming from a branch. https://github.com/nextflow-io/training/pull/512 should be good to merge.

PR #514, opened 2025-02-19
Minor fixes to parts 3, 4, and 5 of Hello Nextflow

mribeirodantas:
- **Fix capitalization of GitHub and Gitpod**
- **Minor fixes to Part 3**
- **Part 4** is very good, but a bit short.
- **Minor fixes to Part 5**

### DIFF
--- a/docs/hello_nextflow/03_hello_workflow.md
+++ b/docs/hello_nextflow/03_hello_workflow.md
@@ -521,7 +521,7 @@ This time the third step was only called once!
 Looking at the output of the `view()` statements, we see the following:
 
 - Three `Before collect:` statements, one for each greeting: at that point the file paths are individual items in the channel.
-- A single `After collect:` statement: the three file paths are now packaged into a singl element.
+- A single `After collect:` statement: the three file paths are now packaged into a single element.
 
 Have a look at the contents of the final output file too:
 
--- a/docs/hello_nextflow/03_hello_workflow.md
+++ b/docs/hello_nextflow/03_hello_workflow.md
@@ -368,7 +368,7 @@ In theory this should handle any arbitrary number of input files.
 
     Some command-line tools require providing an argument (like `-input`) for each input file.
     In that case, we would have to do a little bit of extra work to compose the command.
-    You can see an example of this in the [Nextflow for Genomics](../nf4_science/genomics/) training course.
+    You can see an example of this in the [Nextflow for Genomics](../../nf4_science/genomics/) training course.
 
 <!--[ADD LINK to note above] -->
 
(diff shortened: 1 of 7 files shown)


docs/hello_nextflow/03_hello_workflow.md
```diff
@@ -368,7 +368,7 @@ In theory this should handle any arbitrary number of input files.
 
     Some command-line tools require providing an argument (like `-input`) for each input file.
     In that case, we would have to do a little bit of extra work to compose the command.
-    You can see an example of this in the 'Nextflow for Genomics' training course.
+    You can see an example of this in the [Nextflow for Genomics](https://training.nextflow.io/latest/nf4_science/genomics/) training course.
```
ewels, line 371: Better to use a relative link?

ewels (approved):
LGTM from a quick scan on my phone

docs/hello_nextflow/03_hello_workflow.md
```diff
@@ -521,22 +521,22 @@ This time the third step was only called once!
...
 ```
 
-This time we have all three greetings in the final output file. Success!
+This time we have all three greetings in the final output file. Success! Remove the optional view calls to make the next outputs less verbose.
 
 !!! note
 
     If you run this several times without `-resume`, you will see that the order of the greetings changes from one run to the next.
-    This shows you that the order in which items flow through the pipeline is not guaranteed to be consistent.
+    This shows you that the order in which elements flow through process calls is not guaranteed to be consistent, unless you make use of fair threading (check the [fair](https://www.nextflow.io/docs/latest/reference/process.html#fair) directive).
```
vdauwera, line 539: I don't think the mention of fair threading is helpful, it gives people too much too think about that they don't need at this stage. Let's leave that out.

docs/hello_nextflow/03_hello_workflow.md
```diff
@@ -793,7 +793,7 @@ But as the saying goes, why not both?
 
 ### 4.2. Report the output at the end of the workflow
 
-Now that we have two outputs coming out of the `collectGreetings` process, the `collectGreetings.out` output channel contains two 'tracks':
+Now that we have two outputs coming out of the `collectGreetings` process, the `collectGreetings.out` output multi-channel contains two channels:
```
vdauwera, line 796: Oh yes this is a thing i misunderstood originally, good catch. But I would delete 'multi-channel' because it just complicates the sentence. "the output contains two channels" is clearer.

vdauwera (approved):
All good catches, just a couple of minor change requests to keep things simple. Feel free to merge once you've made the changes.

docs/hello_nextflow/03_hello_workflow.md
vdauwera: yes relative link would be better

docs/hello_nextflow/03_hello_workflow.md
mribeirodantas: Ok, fair point 😄

PR #534, opened 2025-03-03
Add modules component to 'Nextflow for Genomics'

pinin4fjords:
Added a modules component to 'Nexflow for Genomics', needed before adding an nf-test component.

I've intentionally not belaboured the points to much- the basic concepts are now in the new 'Hello Modules' and we don't want to maintain duplicated content in two places.


adamrtalbot (approved):
nailed it

pinin4fjords:
Thanks @adamrtalbot !

PR #538, opened 2025-03-05
Training small fixes

ewels:
Some minor fixes for things that I spotted whilst recording the new training. I made these edits locally before recording, so they are included in the video footage already.

The main change was removing use of the implicit `$it` variable. See https://nextflow.io/docs/latest/updating-syntax.html#deprecated-syntax for details:

CleanShot 2025-03-05 at 09 44 48@2x


pinin4fjords (dismissed):
LGTM. I need to remember to stop doing implicit `it`

PR #539, opened 2025-03-05
Local devcontainers enablement

pinin4fjords:
Two devcontainers files, one for Codespaces, one for local. The codespaces one is unchanged from master, I've just moved it to the subdir.

non-codespaces ('local') operation involves mirroring paths from host, and symlinking to create the /workspaces location for harmony with the materials. We then set all terminals to open in that path.

Symlinked the codespaces version at the root so it's the default.

Local operation would be a little more clunky, but would now work. I've checked that the DoD functionality is restored.


.devcontainer/codespaces/Dockerfile
ewels, line 1: Do you need these symlinks? Can'y you just point to a `Dockerfile` in the directory above? (assuming that this is when uncommenting out `build`?)

.devcontainer/devcontainer.json
ewels, line 1: What's this one for?

ewels (commented):
Diff between the two files for my own benefit:

```diff
2c2
<     "name": "nextflow-training (codespaces)",
---
>     "name": "nextflow-training (local)",
14,15c14,15
<     "workspaceFolder": "/workspaces/training",
<     "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/training,type=bind", // Used for local devcontainers, ignored in Codespaces.
---
>     "workspaceFolder": "${localWorkspaceFolder}",
>     "workspaceMount": "source=${localWorkspaceFolder},target=${localWorkspaceFolder},type=bind",
17,18c17,18
<         "NXF_HOME": "/workspaces/.nextflow",
<         "HOST_PROJECT_PATH": "/workspaces/training" // Codespaces and local devcontainers will both work with this.
---
>         "NXF_HOME": "/workspaces/training/.nextflow",
>         "HOST_PROJECT_PATH": "/workspaces/training"
20,21c20
<     "postCreateCommand": "nextflow -version; if [ -z \"$CODESPACES\" ]; then echo \"Devcontainers Development\"; else echo \"Codespaces Development\";  fi",
<     // Configure tool-specific properties.
---
>     "postCreateCommand": "sudo mkdir -p /workspaces && sudo ln -s ${localWorkspaceFolder} /workspaces/training && if [ -z \"$CODESPACES\" ]; then echo \"Local Development\"; else echo \"Codespaces Development\"; fi",
27c26,28
<                 "python.defaultInterpreterPath": "/opt/conda/bin/python"
---
>                 "python.defaultInterpreterPath": "/opt/conda/bin/python",
>                 "terminal.integrated.cwd": "/workspaces/training",
>                 "terminal.integrated.defaultLocation": "workspace"

```

.devcontainer/codespaces/devcontainer.json
```diff
@@ -0,0 +1,41 @@
...
+    // Uncomment this and comment "image" when testing Dockerfile changes
+    // "build": { "dockerfile": "Dockerfile", "context": ".." },
+
+    "features": {
+        "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
+            "version": "latest"
+        }
+    },
+    "workspaceFolder": "/workspaces/training",
+    "workspaceMount": "source=${localWorkspaceFolder},target=/workspaces/training,type=bind", // Used for local devcontainers, ignored in Codespaces.
```
ewels, line 15: Can we delete this line?

.devcontainer/codespaces/devcontainer.json
```diff
@@ -0,0 +1,41 @@
...
+    "remoteEnv": {
+        "NXF_HOME": "/workspaces/.nextflow",
+        "HOST_PROJECT_PATH": "/workspaces/training" // Codespaces and local devcontainers will both work with this.
+    },
+    "postCreateCommand": "nextflow -version; if [ -z \"$CODESPACES\" ]; then echo \"Devcontainers Development\"; else echo \"Codespaces Development\";  fi",
```
ewels, line 20: Could we get rid of this now?

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,42 @@
...
+            "version": "latest"
+        }
+    },
+    "workspaceFolder": "${localWorkspaceFolder}",
+    "workspaceMount": "source=${localWorkspaceFolder},target=${localWorkspaceFolder},type=bind",
+    "remoteEnv": {
+        "NXF_HOME": "/workspaces/training/.nextflow",
+        "HOST_PROJECT_PATH": "/workspaces/training"
+    },
+    "postCreateCommand": "sudo mkdir -p /workspaces && sudo ln -s ${localWorkspaceFolder} /workspaces/training && if [ -z \"$CODESPACES\" ]; then echo \"Local Development\"; else echo \"Codespaces Development\"; fi",
```
ewels, line 20: Could we get rid of the `if [ -z \"$CODESPACES\" ]; then echo \"Local Development\"; else echo \"Codespaces Development\"; fi` bit?

ewels:
How does this work when creating a codespace? I'm not well read on the default file paths and similar. Does Codespaces pick the correct one by default?

How does local devcontainer setup know which one to use?

.devcontainer/devcontainer.json
pinin4fjords: Just a symlink so it's the default

.devcontainer/codespaces/devcontainer.json
pinin4fjords: If you like :-)

pinin4fjords:
> How does this work when creating a codespace? I'm not well read on the default file paths and similar. Does Codespaces pick the correct one by default?
> 
> How does local devcontainer setup know which one to use?

I think Codespaces will use the one at the root of .devcontainers by default (which is why I added that symlink). The VSCode plugin asks you to choose, so we just pick the local config when it does.

.devcontainer/codespaces/devcontainer.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    // Configure tool-specific properties.
+    "customizations": {
+        // Configure properties specific to VS Code.
+        "vscode": {
+            // Set *default* container specific settings.json values on container create.
+            "settings": {
+                "python.defaultInterpreterPath": "/opt/conda/bin/python"
+            },
+            // Add the IDs of extensions you want installed when the container is created.
```
ewels, line 27:
```suggestion
        "HOST_PROJECT_PATH": "/workspaces/training"
    },
    "customizations": {
        "vscode": {
            "settings": {
                "python.defaultInterpreterPath": "/opt/conda/bin/python"
            },
```

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,42 @@
...
+    "customizations": {
+        // Configure properties specific to VS Code.
+        "vscode": {
+            // Set *default* container specific settings.json values on container create.
+            "settings": {
+                "python.defaultInterpreterPath": "/opt/conda/bin/python",
+                "terminal.integrated.cwd": "/workspaces/training",
+                "terminal.integrated.defaultLocation": "workspace"
+            },
+            // Add the IDs of extensions you want installed when the container is created.
```
ewels, line 30:
```suggestion
        "vscode": {
            "settings": {
                "python.defaultInterpreterPath": "/opt/conda/bin/python",
                "terminal.integrated.cwd": "/workspaces/training",
                "terminal.integrated.defaultLocation": "workspace"
            },
```

ewels (commented):
Remove superfluous comments

ewels:
Updated diff:

```diff
2c2
<     "name": "nextflow-training (codespaces)",
---
>     "name": "nextflow-training (local)",
14c14,15
<     "workspaceFolder": "/workspaces/training",
---
>     "workspaceFolder": "${localWorkspaceFolder}",
>     "workspaceMount": "source=${localWorkspaceFolder},target=${localWorkspaceFolder},type=bind",
16c17
<         "NXF_HOME": "/workspaces/.nextflow",
---
>         "NXF_HOME": "/workspaces/training/.nextflow",
18a20
>     "postCreateCommand": "sudo mkdir -p /workspaces && sudo rm -rf /workspaces/training && sudo ln -s ${localWorkspaceFolder} /workspaces/training",
22c24,26
<                 "python.defaultInterpreterPath": "/opt/conda/bin/python"
---
>                 "python.defaultInterpreterPath": "/opt/conda/bin/python",
>                 "terminal.integrated.cwd": "/workspaces/training",
>                 "terminal.integrated.defaultLocation": "workspace"

```

ewels:
> I think Codespaces will use the one at the root of .devcontainers by default (which is why I added that symlink). The VSCode plugin asks you to choose, so we just pick the local config when it does.

ok yup, agree that [this seems to be the case](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers#default-configuration-selection-during-codespace-creation). I think we could probably pre-set this in the codespaces link that we have in the training, but if that link isn't used then it sounds like a very standard one would be set by default, so safer to symlink it as you did. All good 👍🏻

.devcontainer/codespaces/devcontainer.json
```diff
@@ -0,0 +1,35 @@
...
+            "settings": {
+                "python.defaultInterpreterPath": "/opt/conda/bin/python"
+            },
+            "extensions": [
+                "ms-python.python",
+                "ms-python.vscode-pylance",
+                "nf-core.nf-core-extensionpack", // nf-core recommended extensions
+                "nextflow.nextflow", // Nextflow VS Code extension
+                "codezombiech.gitignore", // Language support for .gitignore files
+                "cssho.vscode-svgviewer" // SVG viewer
```
ewels, line 30: I think we can simplify these - `nextflow.nextflow` shouldn't be needed as it's already part of the nf-core extension pack, and the other two could be added to that. Then we just have the nf-core extension pack and be done with it.

I'd also like to remove a couple of the more subjective packages from the nf-core extension pack (rainbow indentation, TODO tree, code spell checker). I find myself disabling these each time in devcontainers, which is annoying.

We can do this in a separate PR, as needs upstream changes to the nf-core extension pack first.

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,39 @@
...
+                "terminal.integrated.cwd": "/workspaces/training",
+                "terminal.integrated.defaultLocation": "workspace"
+            },
+            "extensions": [
+                "ms-python.python",
+                "ms-python.vscode-pylance",
+                "nf-core.nf-core-extensionpack", // nf-core recommended extensions
+                "nextflow.nextflow", // Nextflow VS Code extension
+                "codezombiech.gitignore", // Language support for .gitignore files
+                "cssho.vscode-svgviewer" // SVG viewer
```
ewels, line 34:
```suggestion
                "codezombiech.gitignore" // Language support for .gitignore files
```

.devcontainer/codespaces/devcontainer.json
ewels, line 30:
```suggestion
                "codezombiech.gitignore" // Language support for .gitignore files
```

ewels (commented):
The SVG viewer extension is archived and no longer present in the marketplace, this can be removed.

.devcontainer/codespaces/devcontainer.json
ewels: https://github.com/nf-core/vscode-extensionpack/pull/25

.devcontainer/codespaces/devcontainer.json
```diff
@@ -0,0 +1,34 @@
+{
+    "name": "nextflow-training (codespaces)",
+
+    // Update this to pin the release tag before release
+    "image": "ghcr.io/nextflow-io/training:latest",
+    // Uncomment this and comment "image" when testing Dockerfile changes
+    // "build": { "dockerfile": "../Dockerfile", "context": ".." },
```
edmundmiller, line 7: Why are we doing this? Because it has to "build" every time?

Couldn't we utilize prebuilds? https://docs.github.com/en/codespaces/prebuilding-your-codespaces/about-github-codespaces-prebuilds

ewels: Not with a free GitHub org, sadly.

ewels:
@pinin4fjords we should update the orientation docs to match this change, ideally with a screenshot etc.

pinin4fjords:
@ewels in the process of verifying docs, I discovered that my changes weren't compatible with your latest tidying up. 

The devcontainers plugin on my Mac insisted on being an ARM64 whatever I tried, and the devcontainers base image you used is AMD64 only. I couldn't get the pre-built image to work with the docker-outside-docker either, which is probably related.

I switched the image to one that's cross-compatible, and made the local devcontainer revert to building each time. Not sure if, if we update `ghcr.io/nextflow-io/training` based on this it will work. If so we can revert the local config to that static image.

ewels:
@pinin4fjords I'd rather just build the image in multiple formats. It's already using buildx, so it shouldn't be difficult to make it multi-arch.

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,39 @@
+{
+    "name": "nextflow-training (local)",
+
+    "build": {
+        "dockerfile": "../Dockerfile",
+        "context": "..",
+        "options": ["--platform=linux/amd64"]
```
ewels, line 7: Do we really want to force people to use amd64? Is this needed?

.devcontainer/local/devcontainer.json
```diff
@@ -0,0 +1,39 @@
...
+                "python.defaultInterpreterPath": "/opt/conda/bin/python",
+                "terminal.integrated.cwd": "/workspaces/training",
+                "terminal.integrated.defaultLocation": "workspace"
+            },
+            "extensions": [
+                "ms-python.python",
+                "ms-python.vscode-pylance",
+                "nf-core.nf-core-extensionpack", // nf-core recommended extensions
+                "nextflow.nextflow", // Nextflow VS Code extension
+                "codezombiech.gitignore" // Language support for .gitignore files
```
ewels, line 34:
```suggestion
```

.devcontainer/codespaces/devcontainer.json
```diff
@@ -0,0 +1,34 @@
...
+        "vscode": {
+            "settings": {
+                "python.defaultInterpreterPath": "/opt/conda/bin/python"
+            },
+            "extensions": [
+                "ms-python.python",
+                "ms-python.vscode-pylance",
+                "nf-core.nf-core-extensionpack", // nf-core recommended extensions
+                "nextflow.nextflow", // Nextflow VS Code extension
+                "codezombiech.gitignore" // Language support for .gitignore files
```
ewels, line 29:
```suggestion
```

ewels (commented):
Merged https://github.com/nf-core/vscode-extensionpack/pull/25

.devcontainer/local/devcontainer.json
pinin4fjords: It actually was. I got some weird Docker shim error without it.

pinin4fjords: But maybe if you build the arm image it won't be needed

ewels: https://github.com/nextflow-io/training/pull/549

PR #547, opened 2025-03-08
Add missing pages to navigation for hello nextflow

ewels:


ewels:
Self merging, as I'm 99% certain it's meant to be like this, and I want to screen record it.

PR #556, opened 2025-03-09
Hide announcement with JS if date is past

ewels:
Use some Javascript to automatically hide the announcement banner if a specified date has passed.

Means that we can still have the banner active even in fixed statically built releases.

Downside is that there is a slight flicker on page load for the banner, I can't get around that though. I've prioritised the flicker for when it's active, vs. when it's inactive (figure it'll be inactive more of the time).


vdauwera (approved):
LGTM

PR #558, opened 2025-03-10
Add docker runOptions to fix default root ownership issues

maxulysse:
I think we might need to add docker runOptions, this seems to be an issue. cf: https://nextflow.slack.com/archives/C02TH6KNBCN/p1741604416453299

### DIFF
--- a/nf4-science/genomics/nextflow.config
+++ b/nf4-science/genomics/nextflow.config
@@ -1,1 +1,2 @@
 docker.enabled = true
+docker.runOptions = '-u $(id -u):$(id -g)'


ewels:
I'd like to look into how https://github.com/nextflow-io/training/pull/539 affects this first. Might be that we can have a separate devcontainers config for local development that doesn't use root, for example?

Might still need this PR as well, but either way I think that we should wait until the above is merged first for testing.

vdauwera:
Is this still a thing or has it been addressed by https://github.com/nextflow-io/training/pull/539 , @ewels ?

ewels:
I don't know any more sorry. I'd recommend chasing back to that slack comment and seeing if the problem can be reproduced.

My guess is that we can close this without doing anything now.

@pinin4fjords is the master of local devcontainers so I also defer to him.

ewels:
Closing for now. Can always reopen if this comes back.

PR #559, opened 2025-03-10
add missing trailing periods

mashehu:
For consistency's sake. (Although they look off anyway (too LaTeXy), imo)


ewels (approved):
Thanks! I want to add a CI check for this, plus correct numbering..

PR #560, opened 2025-03-10
Primary buttons: black text

ewels:
Contrast ratio was poor: https://www.siegemedia.com/contrast-ratio#%23ffffff-on-rgb%2888%2C%20189%2C%20159%29

Black text is better.

For @mashehu 😅


ewels:
Alternative PR to change this colour site-wide (including navbar) here: https://github.com/nextflow-io/training/pull/572

PR #563, opened 2025-03-12
Update orientation page for Side Quests

vdauwera:
Also add it to the nav, update the SQ index and retitle the nf-core SQ.

(Plus update the mkdocs section on header numbers which had repeated and redundant elements.)


vdauwera:
Bypassing review to get this into tonight's release. All very minor changes.

PR #564, opened 2025-03-13
minor numeration fix in in `hello_nextflow/06_hello_config.md`

juanesarango:
Screenshot 2025-03-12 at 11 13 17 PM


vdauwera (approved):
Thanks for the fix!

PR #565, opened 2025-03-13
`nextflow format hello-nextflow/`

ewels:
Ran the new upcoming `nextflow format` command on the `hello-nextflow` directory.
Once this is out, we'll start to run this formatter on a lot of nf-core code and it should quickly become the established formatting style for Nextflow code. I figure it's good to get ahead of the game and get the training to follow this style too.

No manual changes here at all, purely the automated formatting changes. We'll need to update the markdown code to reflect these changes.

Note that in other docs I've used mkdocs material "snippets" to directly include code files into the markdown without copying + pasting, which makes these kinds of changes easier, not sure if it's worth considering here as we update or not worth the effort.. See [docs](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown-extensions/#snippets), [docs](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/) and example:

https://github.com/nextflow-io/training/blob/ce85956786801eb5cdf5df050bd27cf9eb1157dd/docs/basic_training/containers.md?plain=1#L447-L449

Once we have sign off on the use of this formatting tool and the code style it produces, I can update this PR to also adjust the markdown.

Once that's done we should apply it to all other material, then set up pre-commit to automatically check (/ apply on comment) this auto-formatting.

Phil

### DIFF
--- a/hello-nextflow/hello-channels.nf
+++ b/hello-nextflow/hello-channels.nf
@@ -1,24 +1,4 @@
 #!/usr/bin/env nextflow
-
-/*
- * Use echo to print 'Hello World!' to a file
- */
-process sayHello {
-
-    publishDir 'results', mode: 'copy'
-
-    input:
-        val greeting
-
-    output:
-        path 'output.txt'
-
-    script:
-    """
-    echo '$greeting' > output.txt
-    """
-}
-
 /*
  * Pipeline parameters
  */
@@ -28,4 +8,24 @@ workflow {
 
     // emit a greeting
     sayHello(params.greeting)
+}
+
+
+/*
+ * Use echo to print 'Hello World!' to a file
+ */
+process sayHello {
+
+    publishDir 'results', mode: 'copy'
+
+    input:
+    val greeting
+
+    output:
+    path 'output.txt'
+
+    script:
+    """
+    echo '${greeting}' > output.txt
+    """
 }
--- a/hello-nextflow/hello-config.nf
+++ b/hello-nextflow/hello-config.nf
@@ -1,4 +1,10 @@
 #!/usr/bin/env nextflow
+// Include modules
+include { sayHello } from './modules/sayHello.nf'
+include { convertToUpper } from './modules/convertToUpper.nf'
+include { collectGreetings } from './modules/collectGreetings.nf'
+include { cowpy } from './modules/cowpy.nf'
+
 
 /*
  * Pipeline parameters
@@ -7,18 +13,13 @@ params.greeting = 'greetings.csv'
 params.batch = 'test-batch'
 params.character = 'turkey'
 
-// Include modules
-include { sayHello } from './modules/sayHello.nf'
-include { convertToUpper } from './modules/convertToUpper.nf'
-include { collectGreetings } from './modules/collectGreetings.nf'
-include { cowpy } from './modules/cowpy.nf'
-
 workflow {
 
     // create a channel for inputs from a CSV file
-    greeting_ch = Channel.fromPath(params.greeting)
-                        .splitCsv()
-                        .map { line -> line[0] }
+    greeting_ch = Channel
+        .fromPath(params.greeting)
+        .splitCsv()
+        .map { line -> line[0] }
 
     // emit a greeting
     sayHello(greeting_ch)
@@ -30,7 +31,7 @@ workflow {
     collectGreetings(convertToUpper.out.collect(), params.batch)
 
     // emit a message about the size of the batch
-    collectGreetings.out.count.view { "There were $it greetings in this batch" }
+    collectGreetings.out.count.view { "There were ${it} greetings in this batch" }
 
     // generate ASCII art of the greetings with cowpy
     cowpy(collectGreetings.out.outfile, params.character)
--- a/hello-nextflow/hello-workflow.nf
+++ b/hello-nextflow/hello-workflow.nf
@@ -1,4 +1,21 @@
 #!/usr/bin/env nextflow
+/*
+ * Pipeline parameters
+ */
+params.greeting = 'greetings.csv'
+
+workflow {
+
+    // create a channel for inputs from a CSV file
+    greeting_ch = Channel
+        .fromPath(params.greeting)
+        .splitCsv()
+        .map { line -> line[0] }
+
+    // emit a greeting
+    sayHello(greeting_ch)
+}
+
 
 /*
  * Use echo to print 'Hello World!' to a file
@@ -8,29 +25,13 @@ process sayHello {
     publishDir 'results', mode: 'copy'
 
     input:
-        val greeting
+    val greeting
 
     output:
-        path "${greeting}-output.txt"
+    path "${greeting}-output.txt"
 
     script:
     """
-    echo '$greeting' > '$greeting-output.txt'
+    echo '${greeting}' > '${greeting}-output.txt'
     """
-}
-
-/*
- * Pipeline parameters
- */
-params.greeting = 'greetings.csv'
-
-workflow {
-
-    // create a channel for inputs from a CSV file
-    greeting_ch = Channel.fromPath(params.greeting)
-                        .splitCsv()
-                        .map { line -> line[0] }
-
-    // emit a greeting
-    sayHello(greeting_ch)
 }
--- a/hello-nextflow/solutions/1-hello-world/hello-world-3.nf
+++ b/hello-nextflow/solutions/1-hello-world/hello-world-3.nf
@@ -1,4 +1,10 @@
 #!/usr/bin/env nextflow
+workflow {
+
+    // emit a greeting
+    sayHello()
+}
+
 
 /*
  * Use echo to print 'Hello World!' to a file
@@ -8,16 +14,10 @@ process sayHello {
     publishDir 'results', mode: 'copy'
 
     output:
-        path 'output.txt'
+    path 'output.txt'
 
     script:
     """
     echo 'Hello World!' > output.txt
     """
-}
-
-workflow {
-
-    // emit a greeting
-    sayHello()
 }
--- a/hello-nextflow/hello-containers.nf
+++ b/hello-nextflow/hello-containers.nf
@@ -1,4 +1,9 @@
 #!/usr/bin/env nextflow
+// Include modules
+include { sayHello } from './modules/sayHello.nf'
+include { convertToUpper } from './modules/convertToUpper.nf'
+include { collectGreetings } from './modules/collectGreetings.nf'
+
 
 /*
  * Pipeline parameters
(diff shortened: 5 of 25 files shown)


bentsherman:
Please let me know if you have opinions about formatting. I will hear you out. Get your views in before things become set in stone!

hello-nextflow/hello-channels.nf
```diff
@@ -8,24 +19,13 @@ process sayHello {
...
+    val greeting
 
     output:
-        path 'output.txt'
+    path 'output.txt'
 
     script:
     """
-    echo '$greeting' > output.txt
+    echo '${greeting}' > output.txt
```
vdauwera, line 29: are the curly braces going to be mandatory? why do we need them?

hello-nextflow/hello-config.nf
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env nextflow
+// Include modules
```
vdauwera, line 2: Is the idea here that module includes should go before anything else? Is there a functional reason or is it purely stylistic? Any interactions wrt setting any variable values?

hello-nextflow/hello-config.nf
```diff
@@ -7,18 +13,13 @@ params.greeting = 'greetings.csv'
...
-include { collectGreetings } from './modules/collectGreetings.nf'
-include { cowpy } from './modules/cowpy.nf'
-
 workflow {
 
     // create a channel for inputs from a CSV file
-    greeting_ch = Channel.fromPath(params.greeting)
-                        .splitCsv()
-                        .map { line -> line[0] }
+    greeting_ch = Channel
```
vdauwera, line 19: I'm not sure I like having to put the `.fromPath()` bit on the next line. I would keep that like it was before tbh. That way you have the first line do a full thing, then the next lines modify that thing. If you have just `greeting_ch = Channel` on one line that feels... interrupted

hello-nextflow/hello-config.nf
```diff
@@ -30,7 +31,7 @@ workflow {
     collectGreetings(convertToUpper.out.collect(), params.batch)
 
     // emit a message about the size of the batch
-    collectGreetings.out.count.view { "There were $it greetings in this batch" }
+    collectGreetings.out.count.view { "There were ${it} greetings in this batch" }
```
vdauwera, line 34: are the extra curlies necessary here?

vdauwera: also should we be explicitly naming the variable like we do later with `line -> line[0]` etc?

vdauwera (commented):
Added some questions/comments that may be more for @bentsherman

vdauwera:
@ewels Before this can go in we're also going to have to update a bunch of line numbers, check the explanatory text for any code that is quoted and needs to be updated to match etc., so this will still need some substantial effort.

Also, do you have any thoughts on how this will make the code different in a lot of places compared to what you show in the videos? I know we can tell people to expect minor changes but a few of these aren't really minor, at least for newcomers (eg chunks of code that move to a different part of the file) and in general it's going to add 'noise' in terms of people trying to follow what's going on on screen and on their own machine. 

So I'm a little reluctant to push this out right away if there's not a driving need for it. At some level I'd rather sit on it until we feel we may need to update the recordings themselves (hopefully not for some number of months as they are so good and it takes so much effort to make), OR there's a strong reason for why we need the formatting to be strictly aligned.

ewels:
Yup, totally agree - the material will need some significant changes.  Also reordering some sections, eg. introducing workflows first then processes.

I wanted to get feedback on the formatter's style choices first really, to see how many changes it would make and see if we agree with them. Then we can adjust it before it goes live in Nextflow.

Agree that we shouldn't rush this, and happy to leave it for quite a while before doing.

That said, maybe we can progressively implement it, using it for any new courses / material from now on. And maybe applying to the courses that don't have videos. Does that sound ok?

hello-nextflow/hello-channels.nf
bentsherman: They are optional in certain cases like simple variable names. The formatter doesn't try to be smart here, it just always uses braces

hello-nextflow/hello-config.nf
bentsherman: In principle, it's just a convention as it shouldn't matter where you put the includes

In practice, users sometimes put the params first so that they are propagated to the included modules as a side effect. We are trying to move away from that pattern anyway

hello-nextflow/hello-config.nf
bentsherman: See my point above about curly braces. As for the implicit `it`, yes I would recommend as a best practice that you declare an explicit name such as `line`. Right now it's only a "paranoid" warning but it will eventually become an error

hello-nextflow/hello-channels.nf
bentsherman: As for whether we should put them everywhere, there is a trade-off between consistency and shortcuts

Using curly braces everywhere requires a bit more typing but makes the code more consistent, therefore easier to read

Omitting the curly braces where possible saves a bit on typing but makes the code less consistent, thus the reader has to work a bit more to understand the different syntax forms

So I think it comes down to how much time we spend reading vs writing code. At least in software engineering, we spend way more time reading code, so I tend to favor consistency

hello-nextflow/solutions/1-hello-world/hello-world-3.nf
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env nextflow
+workflow {
+
+    // emit a greeting
+    sayHello()
+}
+
```
adamrtalbot, line 7: I dislike having workflow above processes. I use this order:

1. include
2. params
3. functions
4. processes
5. workflows
6. anonymous workflow

bentsherman: I'm debating whether to add an option to control the order (e.g. "Formatting > Entry workflow position")

Or maybe just not try to re-order the definitions at all?

hello-nextflow/hello-config.nf
ewels: For me I would do this for any old `foo.bar()` as well I think.. The first line should contain the first action. Whether it's `Channel` or `foo`, neither of these are actions, only the starting points.

hello-nextflow/hello-config.nf
ewels: I thought I got all of the `$it`s in https://github.com/nextflow-io/training/pull/538 - must have missed this one.

hello-nextflow/solutions/1-hello-world/hello-world-3.nf
ewels: I've always done processes first then workflows, but I'm coming around to it.

I don't mind making @adamrtalbot feel uncomfortable, auto-formatters have that effect on everyone 😅  Though maybe this goes beyond what most formatters do?

adamrtalbot: I'll survive if it gets reordered, but do we have a particular order that is preferred? It feels weird to be opinionated about something that isn't consequential (not that it's stopped me before).

ewels: Isn't that what auto-formatters do? 😆  Basically no formatting is consequential. It's mostly for convention and consistency, so that folks get used to navigating pipeline code in the same way everywhere.

_Any colour, as long as it's black._

adamrtalbot: Not sure I agree - a Python or Go formatter doesn't move classes and functions around. I'm not certain about Java/Groovy autoformatters which are probably the most relevant but from recollection they don't move classes around.

Formatters generally stick to following the style guidelines of the language because these keep the code consistent between projects and make it easier for developers to get started on an existing project.

> It's mostly for convention and consistency, so that folks get used to navigating pipeline code in the same way everywhere.

So the question is - what is the standard here? Because I see processes above workflows far more than I see workflows then processes, but I might be biased.

ewels: Python formatters do often move `import` statements to the top, but yeah I agree that's the exception rather than the rule. But those languages don't really have an equivalent structure I think - Python doesn't have the same watertight assumption that a file will always have these 3 very distinct things.

I also see processes first, I think that your suggested order is a lot less controversial to be honest. I think it was @bentsherman and @vdauwera who convinced me that it made more sense to have workflows first, but yeah if we want to go ahead with this then we will be fighting uphill a bit and need to justify that.

adamrtalbot: > Python formatters do often move import statements to the top

Because that's PEP8 right? So maybe we're dancing around the real issue here - do we have style guidelines for Nextflow? Once we have them the formatter can be as opinionated as Black.

> Any colour, as long as it's black.

heh, nice pun.

hello-nextflow/hello-config.nf
bentsherman: When I imagine something like this:

bentsherman: When I imagine something like this:

```groovy
ch_samples
  .filter { sample -> sample.n_reads >= 1_000_000 }
  .map { sample -> sample.id }
```

I think I prefer both calls to be on their own line. I think I treat channel factories differently in my head because the `Channel` and the factory call are sort of a unit, but in general it seems like the "starting point" should be separate from any transformations performed on it

hello-nextflow/solutions/1-hello-world/hello-world-3.nf
bentsherman: Most languages don't have this problem because they just have functions, no processes / workflows.

I do think that most people put their workflows at the bottom. This is also the convention that Paolo used when he introduced DSL2 in the docs, which could explain it.

I'm starting to think the formatter should just preserve the user's order for now

hello-nextflow/hello-channels.nf
kenibrewer: I think it's easier to teach consistency as well. So I like this change.

hello-nextflow/solutions/1-hello-world/hello-world-3.nf
ewels: Agreed - or at least make it opt in, off by default for now.

ewels: ..Maybe with includes as an exception? I do like having them at the top and alphabetically sorted 👀

hello-nextflow/hello-config.nf
bentsherman: the `format` command will now put `Channel.fromPath()` on a single line, as well as any other channel factory

hello-nextflow/hello-workflow.nf
```diff
@@ -1,4 +1,21 @@
...
+
+    // create a channel for inputs from a CSV file
+    greeting_ch = Channel
+        .fromPath(params.greeting)
+        .splitCsv()
+        .map { line -> line[0] }
+
+    // emit a greeting
+    sayHello(greeting_ch)
+}
```
bentsherman, line 17: the `format` command now will not try to re-order your declarations unless you enable it, should make the formatting less intrusive overall

hello-nextflow/solutions/1-hello-world/hello-world-3.nf
bentsherman: the `format` command now will not try to re-order your declarations unless you enable it, should make the formatting less intrusive overall

ewels:
Closing in favour of the (much less intrusive) https://github.com/nextflow-io/training/pull/787

Thanks for the discussion everyone, and thanks for the formatter updates @bentsherman 👏🏻

PR #568, opened 2025-03-18
refactor: switch to slim devcontainer with features

kenibrewer:
This removes the dockerfile and docker build workflow which doesn't seem to being pulled by codespaces anyways.  Instead we use a base image and devcontainer


kenibrewer:
Closing this in favor of #569

PR #566, opened 2025-03-18
Codespaces: use ubuntu base image, much smaller

ewels:
Testing now..

PR #569, opened 2025-03-18
Codespaces: No Dockerfile

ewels:
@gwright99 shared [this video](https://www.youtube.com/watch?v=M21loGvplVM) with me (see [associated blog](https://containers.dev/guide/prebuild)) which is about pre-builds for devcontainers (NB: not _GitHub prebuilds_ which are a whole thing, but just pre-building docker images for codespaces).

We were already sort of doing some of this, but watching the video and checking the [associated reference repo](https://github.com/craiglpeters/kubernetes-devcontainer) was super useful. The main things being:

1. A lot of what we're doing in the `Dockerfile` can be done with [devcontainer features](https://containers.dev/features) - for example: conda, java. This is much simpler.
2. Other stuff can be done in custom _local_ devcontainer features

By using devcontainer features and install scripts, we do everything we need and don't need a `Dockerfile` at all!

I then went on to follow the guidelines to push a devcontainers image and use that, which should be faster. The first time I tested build / pull side by side, the build was actually fractionally faster 🤯 - it seems that GitHub codespaces are veerrry slow at extracting the docker image layers. The second one was faster with the pre-built image though, and that was the one I recorded:

https://github.com/user-attachments/assets/b8050789-4f3c-49dd-b652-4ce1dcdc1a43

Pre-built is on the left, building from scratch is on the right.

One advantage of _not_ prebuilding is that the devcontainers image would then be effectively pinned with the training version. I haven't figured out how to do this in a good way with the pre-built image, so currently it just uses `:latest`. This means that the codespaces dev environment may not be in sync with the training material version (though the files - the checked out repo - should be fine).

I'm not sure how big a deal this is. Maybe it's better to go faster and always up to date with a `:latest` image?
Or better to pay with an extra minute of build time and have pinned devcontainer configuration?


ewels:
Just saw https://github.com/robsyme/features that was referenced in https://github.com/nextflow-io/training/pull/568

We should compare notes, looks like @robsyme might be doing some fancy-pants stuff in there that might be better than just installing Nextflow and nf-test with conda as I'm doing here.. 🤔

ewels:
The pre-built image is now down to 1.12 GB compressed, which is even smaller than the 1.26GB for the one I did in https://github.com/nextflow-io/training/pull/566

https://hub.docker.com/repository/docker/ewels/nextflow-training/tags

ewels:
Also, this PR should create devcontainer images for both amd and arm platforms..

ewels:
That's better!

```console
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
overlay          32G  3.9G   26G  13% /
tmpfs            64M     0   64M   0% /dev
shm              64M     0   64M   0% /dev/shm
/dev/sdb1        44G   18G   24G  43% /tmp
/dev/loop4       32G  3.9G   26G  13% /workspaces
/dev/root        29G   13G   16G  45% /vscode
tmpfs           1.6G  1.2M  1.6G   1% /run/docker-host.sock
tmpfs           3.9G     0  3.9G   0% /proc/acpi
tmpfs           3.9G     0  3.9G   0% /proc/scsi
tmpfs           3.9G     0  3.9G   0% /sys/firmware
```

pinin4fjords (dismissed):
I ran a codespace from the development config, works fine

.github/workflows/devcontainer-build-and-push.yml
```diff
@@ -35,21 +35,28 @@ jobs:
...
-        uses: docker/build-push-action@v6
+      - name: Pre-build dev container image
+        uses: devcontainers/ci@v0.3
         with:
-          file: .devcontainer/Dockerfile
-          push: ${{ github.event_name != 'pull_request' }}
-          tags: ghcr.io/${{ github.repository }}:${{ env.IMAGE_TAG }}
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          configFile: .devcontainer/development/devcontainer.json
```
adamrtalbot, line 56: Interesting, I didn't know this was possible!

.devcontainer/devcontainer.json
adamrtalbot, line 1: No features because it's using the .github/.devcontainer/devcontainer.json for build, right?

adamrtalbot (approved):
OK, took me a while to understand it (as someone not familiiar with devcontainers). If it was something we edited frequently I would suggest we put some documentation in but for now I think lgtm

.devcontainer/development/local-features/conda-packages/install.sh
```diff
@@ -0,0 +1,21 @@
...
+conda update --quiet --yes --all
+
+conda install --quiet --yes --name base \
+    mamba \
+    nextflow \
+    nf-core \
+    nf-test \
+    black \
+    prettier \
+    pre-commit \
```
kenibrewer, line 17: What do you think about installing python tooling via uv tool install like I do in #568 's postCreateCommand.sh

kenibrewer (commented):
Really like the approach, but might want to tweak some details. Is it alright if I make a commit and have Florian be final reviewer so you Europe folks can go to bed ontime?

.devcontainer/devcontainer.json
ewels: Correct, so they're frozen into the `ghcr.io/nextflow-io/training` image already 👍🏻

.devcontainer/development/devcontainer.json
```diff
@@ -0,0 +1,46 @@
...
+        // Nextflow installation version
+        "NXF_HOME": "/workspaces/.nextflow",
+        "NXF_EDGE": "0",
+        "NXF_VER": "24.10.4",
+        // Other env vars
+        "HOST_PROJECT_PATH": "/workspaces/training",
+        "SHELL": "/bin/bash" // Ush bash
+    },
+    "onCreateCommand": "bash .devcontainer/setup.sh",
+    "customizations": {
```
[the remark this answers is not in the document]
ewels: At the moment they're duplicated so present in both. I did that so that people could choose between using the pre-built image or not, both `devcontainer` configs should work in a very comparable manner.

What's the reason to remove them from here?

.github/workflows/devcontainer-build-and-push.yml
```diff
@@ -35,21 +35,28 @@ jobs:
...
+          imageName: ghcr.io/${{ github.repository }}
+          imageTag: ${{ env.IMAGE_TAG }}
```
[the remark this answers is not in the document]
ewels: True, I was building off what was already in the action. Is there much difference between the two approaches? I did think about it when I wrote the env code the other week, but figured it was six of one and half a dozen of the other.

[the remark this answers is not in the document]
adamrtalbot: no, it might be have slightly fewer edge cases, but really not many.

Most important thing would be autocomplete works in your editor.

.devcontainer/development/devcontainer.json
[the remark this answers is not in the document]
kenibrewer: That makes sense. Thanks!

adamrtalbot (approved):

.devcontainer/development/local-features/conda-packages/install.sh
kenibrewer: The postCreateCommand was just to point you to the right part of the code. It only takes a fraction of a second to complete, so it seemed unnecessary to include it in the build. I'm good with it in either place.

We can save the pixi implementation for a later improvement.

kenibrewer (approved):
LGTM

PR #599, opened 2025-04-10
Devcontainers: Move postCreate setup.sh into initialisation steps (and container)

ewels:
Hoping that this will fix the "Cannot find Java" problem currently happening on `master`:

CleanShot 2025-04-10 at 16 37 38@2x

My best guess is that initialisation is slow for some reason, and the JAVA variable is not being `unset` quickly enough.

Unfortunately, this is difficult to test before merging. As the point is that it's baking it into the docker image, which only happens on merge.


ewels:
I have no idea if this works, makes things worse, breaks stuff or what. I seem to get a different critical error every time I try to test it. I'm going to go and have a tantrum and then sulk in my room. Never liked devcontaners anyway..

sulk

pinin4fjords:
Closed in favour of https://github.com/nextflow-io/training/pull/600

PR #602, opened 2025-04-22
Fix for small inconsequential typos

FriederikeHanssen:


docs/side_quests/nf-test.md
```diff
@@ -566,7 +565,7 @@ Learn how to test a Nextflow process.
 
 ---
 
-## 2 Test a Nextflow process
+## 2.0. Test a Nextflow process
```
mribeirodantas, line 568:
```suggestion
## 2.0. Test a Nextflow process
```

Based on the other side quests, with the exception of 1.0 in nf-test, they're all without the 0 in the first section indentations. Maybe we should instead fix the exceptions removing the .0. What do you think?

FriederikeHanssen: Sure. Done in [6d7db0d](https://github.com/nextflow-io/training/pull/602/commits/6d7db0d3b4f7d9425ba4314d76eb595e1303d69b)

ewels: Can add this to the headings CI script if you like?

ewels (approved):

PR #605, opened 2025-04-30
Consistency with using symlink mode

FriederikeHanssen:
In the genomics training we are using symlinks in the beginning, but switch to copy mode in the end.

this PR adds symlink usage throughout. It also adds symlink usage the the rnaseq training for consistency sake


maxulysse (approved):

vdauwera:
Thanks for getting this done!

PR #609, opened 2025-05-10
Install devcontainer tools natively instead of using conda

kenibrewer:
The goal of this PR is to install tools needed by the dev container natively instead of using conda's environment management approach. This approach is faster and less error-prone and is appropriate within a dev container when there is no need for secondary encapsulation of the environment. 

nf-test and nextflow are installed using the bash installers. 
python-based cli tools such as `nf-core` and `pre-commit` are installed using `uv tool install`

I recommend that moving forward we should try to use `conda` only when needed within the context of running nextflow pipelines for the training, and continue managing other installed dependencies for the devcontainer using similar native installation approaches. 

This PR has been tested using both the local devcontainers and codespaces.


vdauwera:
Is there anything I can do to help test this? Fwiw when I start the codespace from the branch deploy preview I get this:

````
training -> which java
/opt/conda/bin/java
training -> java -version
openjdk version "17.0.14-internal" 2025-01-21
OpenJDK Runtime Environment (build 17.0.14-internal+0-adhoc..src)
OpenJDK 64-Bit Server VM (build 17.0.14-internal+0-adhoc..src, mixed mode, sharing)
````

And Nextflow does run, though I haven't tried anything from the trainings yet

vdauwera (approved):
Works in my hands

PR #611, opened 2025-05-16
Metadata Training

FriederikeHanssen:


docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
+# Metadata
+
+Metadata is information that describes and gives context to your data. In workflows, it helps track important details about samples, experimental conditions that can influence processing parameters. Metadata is sample-specific and helps tailor analyses to each dataset's unique characteristics.
```
adamrtalbot, line 3:
```suggestion
Metadata is information that describes and gives context to your data. In workflows, it helps track important details about samples and experimental conditions that can influence processing parameters. Metadata is sample-specific and helps tailor analyses to each dataset's unique characteristics.
```

Maybe not "processing parameters", but I'm not sure what else to put...

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+Think of it like a library catalog: while books contain the actual content (raw data), the catalog cards provide essential information about each book - when it was published, who wrote it, where to find it (metadata). In Nextflow pipelines, metadata helps us:
```
adamrtalbot, line 5:
```suggestion
Think of it like a library catalog: while books contain the actual content (raw data), the catalog cards provide essential information about each book - when it was published, who wrote it, where to find it (metadata). In Nextflow pipelines, metadata can be used to:
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+In the samplesheet, we have both the input files and data about the input files (`id`, `character`), the meta data. As we progress through the workflow, we generate more meta data about each sample. To avoid having to keep track of how many fields we have at any point in time and making the input of our processes more robust, we can combine the meta information into its own key-value paired map.
+
+This separation makes it easier to:
+
+- Track sample information throughout the workflow
+- Add new metadata as you process samples
+- Keep process inputs/outputs clean and organized
+- Query and filter samples based on their properties
+
+Now let's use this and separate our metadata from the file path. We'll use the `map` operator to restructure our channel elements into a tuple consisting of the meta map and file:
```
adamrtalbot, line 164: I think the motivation for this bit isn't very clear. Why do I want to separate the file from the values?

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+[[id:sampleA, character:squirrel], /workspaces/training/side-quests/metadata/data/bonjour.txt]
+[[id:sampleB, character:tux], /workspaces/training/side-quests/metadata/data/guten_tag.txt]
+[[id:sampleC, character:sheep], /workspaces/training/side-quests/metadata/data/hallo.txt]
+[[id:sampleD, character:turkey], /workspaces/training/side-quests/metadata/data/hello.txt]
+[[id:sampleE, character:stegosaurus], /workspaces/training/side-quests/metadata/data/hola.txt]
+[[id:sampleF, character:moose], /workspaces/training/side-quests/metadata/data/salut.txt]
+[[id:sampleG, character:turtle], /workspaces/training/side-quests/metadata/data/ciao.txt]
+```
+
+We have successfully separate our meta data into its own map to keep it next to the file data. Each of our channel elements now has the shape `[ meta,  file ]`.
```
adamrtalbot, line 205:
```suggestion
We have successfully separated our values into their own map to separate it from the file. Each of our channel elements now has the shape `[ meta,  file ]`.
```

Then finish with a sentence saying why this is good.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+### Takeaway
+
+In this section, you've learned:
+
+- **Reading in a samplesheet**: Using `splitCsv` to read CSV files with header information and transform rows into structured data
```
adamrtalbot, line 211:
```suggestion
- **Why metadata is important?** Something here about preserving sample info
- **Reading in a samplesheet**: Using `splitCsv` to read CSV files with header information and transform rows into structured data
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+        ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
+                    .splitCsv(header: true)
+                    .map { row ->
+                        [ [id:row.id, character:row.character], row.recording ]
+                    }
+                    .view()
+    }
+    ```
+
+The tool [langid](https://github.com/saffsd/langid.py) is a language identification tool. It is pre-trained on a set of languages. For a given phrase, it prints a language guess and a probability score for each guess to the console. In the `script` section, we are removing the probability score, clean up the string by removing a newline character and return the language guess. Since it is printed directly to the console, we are using Nextflow's [`stdout` output qualifier](https://www.nextflow.io/docs/latest/process.html#outputs), passing the string on as output.
```
adamrtalbot, line 259:
```suggestion
The tool [langid](https://github.com/saffsd/langid.py) is used for language identification. It comes pre-trained on a set of languages. For a given phrase, it outputs a language prediction and a probability score for each guess to the console. In the `script` section, we remove the probability score, clean up the string by removing newline characters, and return only the language prediction. Since the output is printed directly to the console, we use Nextflow’s [`stdout` output qualifier](https://www.nextflow.io/docs/latest/process.html#outputs) to capture and pass the string as output.
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+This is a useful way to ensure the sample-specific meta information stays connected with any new information that is generated.
+
+### 2.2 Associate the language prediction with the input file
+
+At the moment, our sample files and their language prediction are separated in two different channels: `ch_samplesheet` and `ch_predictions`. But both channels have the same meta information associated with the interesting data points. We can use the meta map to combine our channels back together.
+
+Nextflow includes many methods for combining channels, but in this case the most appropriate operator is [`join`](https://www.nextflow.io/docs/latest/operator.html#join). If you are familiar with SQL, it acts like the `JOIN` operation, where we specify the key to join on and the type of join to perform.
+
+If we check the [`join`](https://www.nextflow.io/docs/latest/operator.html#join) documentation, we can see that it joins two channels based on a defined item, by default the first item in each tuple. If you don't have the console output still available, let's run the pipeline to check our data structures. If you removed the `view()` operator from the `ch_samplesheet` add it back in:
```
adamrtalbot, line 329: Do we want to do this or just pass the file through the process? It might add too much too early.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+```console
+[meta, file, lang]  // e.g. [[id:sampleA, character:squirrel], bonjour.txt, fr]
+```
+
+The `map` operator takes each channel element and processes it to create a modified version. Inside the closure `{ meta, file, lang -> ... }`, we then take the existing `meta` map, create a new map `[lang:lang]`, and merge both together using `+`, Groovy's way of combining maps.
+
+### 2.4 Assign a language group using a ternary operator
+
+Alright, now that we have our language predictions, let's use the information to assign them into new groups. In our example data, we have provided data sets that belong either to `germanic` (either English or German) or `romanic` (French, Spanish, Italian) languages.
```
adamrtalbot, line 527: In English we call them "romance" languages, but we should definitely check.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+      ch_languages = ch_samplesheet.join(ch_prediction)
+                                  .map { meta, file, lang ->
+                                      [ meta + [lang:lang], file ]
+                                  }
+                                  .map{ meta, file ->
+                                      def lang_group = (meta.lang.equals('de') || meta.lang.equals('en')) ? 'germanic' : 'romanic'
+                                      [ meta + [lang_group:lang_group], file ]
+                                  }
+                                  .view()
```
adamrtalbot, line 566: This is a bit messy...can we do something cleaner?

We could either (A) write a function or (B) use case or (C) use good old `if` statements.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+[[id:sampleG, character:turtle, lang:it, lang_group:romanic], /workspaces/training/side-quests/metadata/data/ciao.txt]
+```
+
+Let's understand how this transformation works. The `map` operator takes a closure that processes each element in the channel. Inside the closure, we're using a ternary operator to create a new language group classification.
+
+The ternary expression `(meta.lang.equals('de') || meta.lang.equals('en')) ? 'germanic' : 'romanic'` works like this:
+
+- First, it evaluates the condition before the `?`: checks if the language is either German ('de') or English ('en')
+- If the condition is true (language is German or English), it returns 'germanic'
+- If the condition is false (any other language), it returns 'romanic'
```
adamrtalbot, line 621: Simplifying or removing ternaries will keep this more focused on metadata.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+      `----.          /       }     |        /    \__/
+            `-.      |       /      |       /      `. ,~~|
+                ~-.__|      /_ - ~ ^|      /- _      `..-'
+                     |     /        |     /     ~-.     `-. _  _  _
+                     |_____|        |_____|         ~ - . _ _ _ _ _>
+```
+
+Look through the other files. All phrases should be spoken by the fashionable stegosaurus.
+
+How did this work? The `publishDir` directive is evaluated at runtime when the process executes.Each process task gets its own meta map from the input tuple When the directive is evaluated, `${meta.lang}` is replaced with the actual language value for that sample creating the dynamic paths like `results/fr`.
```
adamrtalbot, line 902:
```suggestion
How did this work? The `publishDir` directive is evaluated at runtime when the process executes. Each process task gets its own meta map from the input tuple When the directive is evaluated, `${meta.lang}` is replaced with the actual language value for that sample creating the dynamic paths like `results/fr`.
```

adamrtalbot (commented):
There's some heavy overlap with splitting and joining, I might be tempted to reduce it and stick with using meta values.

Simplify the ternary stuff, I think it takes attention away from metamaps.

docs/side_quests/metadata.md
pinin4fjords: I agree, this should be a two-step thing:

1) just pass the input file to the output (they've seen that pattern in earlier sections if they've done Genomics)
2) actually, that's a bad plan, because X. What we usually do is keep them separate to keep channel structure simple, yada yada yada

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+    ```groovy title="main.nf" linenums="20" hl_lines="8 11"
+    workflow  {
+
+          ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
+                      .splitCsv(header: true)
+                      .map { row ->
+                          [ [id:row.id, character:row.character], row.recording ]
+                      }
+                      .view()
```
pinin4fjords, line 341:
```suggestion
                      .view{ "samplesheet item: " + it }
```
We should do this sort of thing if there are going to be multiple view ops active at once.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+[[id:sampleA, character:squirrel], /workspaces/training/side-quests/metadata/data/bonjour.txt, fr]
+[[id:sampleB, character:tux], /workspaces/training/side-quests/metadata/data/guten_tag.txt, de]
+[[id:sampleC, character:sheep], /workspaces/training/side-quests/metadata/data/hallo.txt, de]
+[[id:sampleD, character:turkey], /workspaces/training/side-quests/metadata/data/hello.txt, en]
+[[id:sampleE, character:stegosaurus], /workspaces/training/side-quests/metadata/data/hola.txt, es]
+[[id:sampleF, character:moose], /workspaces/training/side-quests/metadata/data/salut.txt, fr]
+[[id:sampleG, character:turtle], /workspaces/training/side-quests/metadata/data/ciao.txt, it]
+```
+
+It is becoming a bit hard to see, but if you look all the way on the right side, you can see that now each of our language predictions is associated with our input files.
```
pinin4fjords, line 445:
```suggestion
It is becoming a bit hard to see, but if you look all the way on the right side, you can see that now each of our language predictions is associated with our input files.
Each channel is not composed of 3 items:

1. The meta map (`[id:sampleA, character:squirrel]`)
2. The input file (`/workspaces/training/side-quests/metadata/data/bonjour.txt`)
3. The language identifier (`fr`)
```

Just clarify, rather than highlighting the difficulty in parsing.

docs/side_quests/metadata.md
pinin4fjords: Definitely romance.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+      ch_prediction = IDENTIFY_LANGUAGE(ch_samplesheet)
+
+      ch_languages = ch_samplesheet.join(ch_prediction)
+                                  .map { meta, file, lang ->
+                                      [ meta + [lang:lang], file ]
+                                  }
+                                  .map{ meta, file ->
+                                      def lang_group = (meta.lang.equals('de') || meta.lang.equals('en')) ? 'germanic' : 'romanic'
+                                      [ meta + [lang_group:lang_group], file ]
+                                  }
```
pinin4fjords, line 565:
```suggestion
                                  .map { meta, file, lang ->
                                      def lang_group = (lang in ['de', 'en']) ? 'germanic' : 'romanic'
                                      [ meta + [ lang: lang, lang_group: lang_group ], file ]
                                  }
```

Single map?

docs/side_quests/metadata.md
pinin4fjords, line 621: I say just scrap this bit. You talk about ternaries above and point to docs, that's sufficient.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+We store this result in the `lang_group` variable and then add it to our meta map using `meta + [lang_group:lang_group]`. The resulting channel elements maintain their `[meta, file]` structure, but the meta map now includes this new classification. This allows us to group samples by their language family later in the workflow.
+
+### Takeaway
+
+In this section, you've learned:
+
+- **Merging on meta maps**: You used `join` to combine two channels based on their meta maps to maintain relationships across processes and channels
+- **Creating custom keys**: You created two new keys in your meta map, adding them with `meta + [new_key:value]`. One based on a computed value from a process, and one based on a condition you set in the `map` operator.
+- **Ternary operator**: You used the ternary operator to determine which language belongs to which group.
```
pinin4fjords, line 631:
```suggestion
In this section, you've learned how to:

- **Merge on meta maps**: You used `join` to combine two channels based on their meta maps to maintain relationships across processes and channels
- **Create custom keys**: You created two new keys in your meta map, adding them with `meta + [new_key:value]`. One based on a computed value from a process, and one based on a condition you set in the `map` operator.
- **Use a ternary operator**: You used the ternary operator to determine which language belongs to which group.
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+## 3. Filter data based on meta map values
+
+We can use the [`filter` operator](https://www.nextflow.io/docs/latest/operator.html#filter) to filter the data based on a condition. Let's say we only want to process romanic language samples further. We can do this by filtering the data based on the `lang_group` field. Let's create a new channel that only contains romanic languages and `view` it:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="20" hl_lines="20-23"
+    workflow  {
+
+      ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
```
pinin4fjords, line 646: You don't need to repeat such big sections, it just detracts from the message

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+- How to use `filter` to select samples based on metadata
+
+We now have only the romanic language samples left and can process those further. Next we want to make characters say the phrases.
+
+---
+
+## 4. Customize a process with meta map
+
+Let's let the characters say the phrases that we have passed in. In the [hello-nextflow training](../hello_nextflow/05_hello_containers.md), you already encountered the `cowpy` package, a python implementation of a tool called `cowsay` that generates ASCII art to display arbitrary text inputs in a fun way. We will re-use a process from there.
```
pinin4fjords, line 735:
```suggestion
Let's let some fun characters say the phrases that we have passed in. In the [hello-nextflow training](../hello_nextflow/05_hello_containers.md), you already encountered the `cowpy` package, a python implementation of a tool called `cowsay` that generates ASCII art to display arbitrary text inputs in a fun way. We will re-use a process from there.
```

'the' implies we know what characters you're referring to, which we don't yet.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+
+This is a subtle difference to other parameters that we have set in the pipelines in previous trainings. A parameter that is passed as part of the `params` object is generally applied to all samples. When a more surgical approache is necessary, using the sample specific information is a good alternative.
+
+### Takeaway
+
+In this section, you've learned:
+
+- **Tweaking directives using meta values**: Using meta map values in `publishDir` directives to create dynamic output paths based on sample properties
+
+- **Tweaking script section based on meta values**: Customizing tool parameters per sample using meta information in the `script` section
```
pinin4fjords, line 958:
```suggestion
In this section, you've learned how to:

- **Tweak directives using meta values**: Using meta map values in `publishDir` directives to create dynamic output paths based on sample properties

- **Tweak the script section based on meta values**: Customizing tool parameters per sample using meta information in the `script` section
```

pinin4fjords (commented):
Nice work great simple subject to teach with! In general:

1. Having the 'after' before the 'before' in examples makes me go a bit cross-eyed
2. You don't need to include such large sections of text in the 'before', just enough to anchor the reader in the text. Copy whole blocks repeatedly is hard to parse, and for us to maintain!
3. It definitely is 'romance' rather than 'romanic'.

docs/side_quests/metadata.md
FriederikeHanssen: removed it

docs/side_quests/metadata.md
FriederikeHanssen: I was first thinking that we could remove it completely, but we need it to make sure the right prediction ends up with the right file. I have been playing around with it, and there is too much stuff happening between this step and the next process to simply try and fail and try again with join. we might be able to show that channels don't have order with a view but it is not guaranteed

docs/side_quests/metadata.md
FriederikeHanssen: good point. You are introducing functions later, don't want to overload this module even more, I am considering just doing a basic if/else. I like ternaries and I think we need to introduce them because they are everywhere, but maybe later

FriederikeHanssen: https://github.com/nextflow-io/training/commit/7f6f29996692299f0810ea4d81efde0b2aa076cf

docs/side_quests/metadata.md
FriederikeHanssen: https://github.com/nextflow-io/training/commit/7f6f29996692299f0810ea4d81efde0b2aa076cf

removed it and instead added a sentence about merging of maps and adding new keys

FriederikeHanssen:
> 1. Having the 'after' before the 'before' in examples makes me go a bit cross-eyed

I kinda agree, although I think in the teaching it works, because people don't need to click to get the "new" code. We are doing it in other places in the training though, but are also inconsistent in how we format it.  Maybe we need an overarching push for consistency here. 
Screenshot 2025-07-16 at 20 30 20
Screenshot 2025-07-16 at 20 30 04
Screenshot 2025-07-16 at 20 27 48

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+                            .splitCsv(header: true)
+
+}
+```
+
+!!! note
+
+    Throughout this tutorial, we'll use the `ch_` prefix for all channel variables to clearly indicate they are Nextflow channels.
+
+=== "After"
```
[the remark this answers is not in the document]
FriederikeHanssen: yes for consistency with other trainings

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+                            .view()
+    ```
+
+=== "Before"
+
+    ```groovy title="main.nf" linenums="2"
+    ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
+    ```
+
+We can use the [`splitCsv` operator](https://www.nextflow.io/docs/latest/operator.html#splitcsv) to split the samplesheet into a channel of maps, where each map represents a row from the CSV file.
```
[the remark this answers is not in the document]
FriederikeHanssen: https://github.com/nextflow-io/training/commit/41f19f31374dc360bceb7bff7a70f3f411036b87

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+[id:sampleA, character:squirrel, recording:/workspaces/training/side-quests/metadata/data/bonjour.txt]
+[id:sampleB, character:tux, recording:/workspaces/training/side-quests/metadata/data/guten_tag.txt]
+[id:sampleC, character:sheep, recording:/workspaces/training/side-quests/metadata/data/hallo.txt]
+[id:sampleD, character:turkey, recording:/workspaces/training/side-quests/metadata/data/hello.txt]
+[id:sampleE, character:stegosaurus, recording:/workspaces/training/side-quests/metadata/data/hola.txt]
+[id:sampleF, character:moose, recording:/workspaces/training/side-quests/metadata/data/salut.txt]
+[id:sampleG, character:turtle, recording:/workspaces/training/side-quests/metadata/data/ciao.txt]
+```
+
+We can see that each row from the CSV file has been converted into a map with keys matching the header row. A map is a key-value data structure similar to dictionaries in Python, objects in JavaScript, or hashes in Ruby. For example:
```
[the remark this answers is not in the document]
FriederikeHanssen: https://github.com/nextflow-io/training/commit/41f19f31374dc360bceb7bff7a70f3f411036b87

docs/side_quests/metadata.md
FriederikeHanssen: removed ternaries completely now, sorry. Love them, but I tend to agree with Adam that it was a bit much even though this training is going to be intermediate.

docs/side_quests/metadata.md
FriederikeHanssen: thought about this more. i think I misunderstood you orginally. I rewrote this now to remove this section entirely and just pass the file through and leave it at that. We can talk about efficient ways to implement this in splitting and grouping I think and just stay focus on meta maps here. 

I assume you meant:

```
      tuple val(meta), path(file), stdout
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,1039 @@
...
+[[id:sampleA, character:squirrel], fr]
+[[id:sampleB, character:tux], de]
+[[id:sampleC, character:sheep], de]
+[[id:sampleD, character:turkey], en]
+[[id:sampleE, character:stegosaurus], es]
+[[id:sampleF, character:moose], fr]
+[[id:sampleG, character:turtle], it]
+```
+
+Neat, for each of our samples, we now have a language predicted. You may have noticed something else: we kept the meta data of our samples and associated it with our new piece of information. We achieved this by adding the `meta` map to the output tuple in the process:
```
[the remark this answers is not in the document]
FriederikeHanssen: Done

docs/side_quests/metadata.md
FriederikeHanssen: removed becaus join was removed

PR #596, opened 2025-04-09
CI: Validate headings

ewels:
New script + GitHub actions workflow to validate any markdown files that have been edited in a PR. Checks:

* Number has a trailing `.`
* Markdown heading depth matches number (eg. `### 1.2.` and not `# 1.2.`)
* Numbers start at 0 or 1 for a given heading depth
* Numbers are sequential

Can be run locally on all files, also has a `--fix` parameter to automagically fix the markdown in-place.

CleanShot 2025-04-09 at 18 14 14@2x

Sometimes it's not always super clear why errors are there, for example `02_hello_modules.md` has:

* `2.`
* `2.1.1.`
* `2.2.`

This throws an error:

```
Numbering must start at 0 or 1  Expected 0 or 1, but got 2.
```

And fix will do something a bit confusing like changing this to:

* `2.`
* `2.1.1.`
* `2.1.`

Also for heading re-numbering fixes, you often have to re-run the script a bunch of times as it goes through fixing subsequent headings.

So handle the diff with some care, probably not ready for a bot auto-fix workflow command yet 😅  

Hopefully helpful though.

PR #617, opened 2025-06-04
Blindly auto-fix all headings

ewels:
There seem to be some dodgy headings that are now causing CI to fail after https://github.com/nextflow-io/training/pull/596 is merged, if those files are touched.

Ran the auto-fix script to clean them up here, but might need some oversight - especially if some of these are affected by drift between English / translated versions.


vdauwera (approved):
There were a few wrong fixes where the right thing to do was ambiguous. Fixed them all as far as I can tell, should be good to go.

vdauwera:
Actually I caught a few more, nobody merge until i finish fixing this

vdauwera (approved):
I think I got them all this time.

PR #630, opened 2025-07-08
Minor edit: add tree depth limit to match output in material

pontushojer:
Just working through the training material (great stuff!!) and noticed that suggested code snipped did not match the expected output here


adamrtalbot (approved):
yep, reproduced and confirmed to fix!

vdauwera:
Good catch, thank you!

PR #634, opened 2025-07-18
Update install.sh

mribeirodantas:
The apptainer `install.sh` script is apparently failing silently. This fix should make Apptainer work again inside the container image.

### DIFF
--- a/.devcontainer/local-features/apptainer/install.sh
+++ b/.devcontainer/local-features/apptainer/install.sh
@@ -1,8 +1,13 @@
 #!/usr/bin/env bash
 
 # Install Apptainer (Singularity)
-add-apt-repository -y ppa:apptainer/ppa
+
 apt-get update --quiet
+
+# installs add-apt-repository
+apt install --reinstall -y software-properties-common
+
+add-apt-repository -y ppa:apptainer/ppa
 
 apt install -y apptainer
 


nathanweeks:
I verified that this PR results in a successful apptainer installation.

However, I ran into the following error when attempting to use apptainer in a codespace (using the "codespaces-dev" dev container, i.e. .devcontainer/codespaces-dev/devcontainer.json) :
```
/tmp -> apptainer pull docker://docker.io/library/ubuntu:24.04
...
/tmp -> apptainer exec ubuntu_24.04.sif cat /etc/os-release
INFO   : A system administrator may need to enable user namespaces, install
INFO   :   apptainer-suid, or compile with ./mconfig --with-suid
ERROR  : Failed to create user namespace: user namespace requires to set /proc/sys/kernel/unprivileged_userns_clone to 1
/tmp -> cat /proc/sys/kernel/unprivileged_userns_clone
1       
```

Ultimately the following dev container runArgs were required to resolve all errors with the above `apptainer exec` command:

```
    "runArgs": [
        "--device=/dev/fuse",
        "--security-opt=seccomp=unconfined",
        "--security-opt=apparmor=unconfined"
    ],
```

For the "local-dev" dev container (.devcontainer/local-dev/devcontainer.json), the following sufficed in my environment---RockyLinux 8.10 with rootless Podman:

```
    "runArgs": [
        "--device=/dev/fuse"
    ],
```
though the other two `--security-opt...` options did no harm. (Note the "docker-outside-of-docker" feature does not appear to be immediately compatible with rootless podman, and was commented out).

vdauwera:
What's the status of this? Is it still needed, and if so what do we need to get it done?

vdauwera:
@mribeirodantas Is this good to go?

mribeirodantas:
Actually, no. Singularity is not even being installed anymore :|. As the only training using singularity is the archived one, I think we can close this with wontfix.

vdauwera:
Dropping this

PR #635, opened 2025-07-20
🧹 cleanup: nextflow language server error fixes

kenibrewer:
Moving some blocks around to reduce the number of strong errors that the Nextflow language server complains about.


ewels:
I think we should just delete this archived material, this problem is only going to get worse.

https://github.com/nextflow-io/training/pull/790

ewels:
Code is gone... 💥

PR #639, opened 2025-07-28
Add debugging side quest content

pinin4fjords:
Adds content for debugging, as part of our new development theme.


docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+### 4.1. Execution Reports and Trace Files
+
+#### Generate Complete Debugging Information
+
+Always run with comprehensive reporting when debugging, either by adding flags to your command line:
+
+```bash
+nextflow run workflow.nf -with-trace -with-timeline -with-report
+````
```
FriederikeHanssen: Formatting is off here

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+Each task creates a work directory with complete debugging information:
+
+```bash
+# Use the hash from trace.txt to find the work directory
+ls -a work/1c/c77799*/
+```
+
+```console title="Work directory contents"
+.               ..              .command.begin  .command.err    .command.log    .command.out    .command.run    .command.sh     .command.trace  .exitcode       [sample2,
+```
```
FriederikeHanssen: maybe we can use `tree` or `ls -lha` to have this fit better onto the screen

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+## What you should know first
+
+This guide assumes you've completed the [Hello Nextflow](../hello_nextflow/) training course and are comfortable with foundational Nextflow concepts including:
+
+- **Basic workflow structure**: Understanding processes, workflows, and how they connect together
+- **Channel operations**: Creating channels, passing data between processes, and using basic operators
+- **Modules and organization**: Creating reusable modules and using include statements
+- **Configuration basics**: Using `nextflow.config` for parameters, process directives, and profiles
+- **Container usage**: Specifying containers for software provisioning
+- **Running workflows**: Executing workflows locally and interpreting basic outputs
```
FriederikeHanssen: more of a general comment on consistency probably for @vdauwera : how do you like us to refer to previous trainings / prerequsites? 

I like this one, but in other trainings we have justlinked out to the trainings: https://training.nextflow.io/latest/side_quests/splitting_and_grouping/#01-prerequisites

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+---
+
+## 1. Syntax Errors
+
+Syntax errors are the most common type of error you'll encounter when writing Nextflow code. They occur when the code does not conform to the expected syntax rules of the Nextflow DSL. These errors prevent your workflow from running at all, so it's important to learn how to identify and fix them quickly.
+
+!!! tip "IDE Features for Syntax Debugging"
+
+    For comprehensive coverage of IDE syntax highlighting, error detection, Problems panel, and other visual debugging aids, see [IDE Features for Nextflow Development](ide_features.md). This section focuses on interpreting error messages and systematic debugging approaches.
```
FriederikeHanssen: Duplicates from above.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+ -- Check '.nextflow.log' file for details
+```
+
+**Key elements of syntax error messages:**
+
+- **File location**: Shows exactly which file contains the error
+- **Line and column**: Points to where the parser encountered the problem
+- **Error description**: Explains what the parser found that it didn't expect
+- **Context**: Shows the problematic line with a caret (^) pointing to location of an unclosed brace
+- **Additional notes**: Provides hints about common causes
```
FriederikeHanssen: you could point / call out where they can see these in the error message

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+    output:
+    path "${sample_name}_output.txt"
+
+    script:
+    """
+    echo "Processing ${sample_name}" > ${sample_name}_output.txt
+    """
+// Missing closing brace for the process
+
```
FriederikeHanssen, line 159: I think for me a bfore/after box with a blue line highlight would be easier to parse. Please ignore, if you disagree. 

It might also be easier to maintain than the screenshots

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+For the purpose of this example we've left a comment for you to show where the error is. The Nextflow VSCode extension should also be giving you some hints about what might be wrong, putting the mismatched brace in red and highlighting the premature end of the file:
+
+![Bad syntax](img/bad_syntax.png)
+
+**Debugging strategy for bracket errors:**
+
+1. Use VS Code's bracket matching (place cursor next to a bracket)
+2. Check the Problems panel for bracket-related messages
+3. Ensure each opening `{` has a corresponding closing `}`
+
```
FriederikeHanssen, line 179: I think we should make the IDE one required for this training actually. It would be hard to find the Problems panel without that training. 

Maybe highlight the error message. I don't know if everyone knows what <EOF> file means and the message is different from the one you get when running the pipeline.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+If you replace the comment with the missing closing brace, and run the workflow again, it should now run successfully:
+
+```bash
+nextflow run bad_syntax.nf
+```
+
+### 1.2. Using incorrect process keywords or directives
+
+Another common syntax error is an **invalid process definition**. This can happen if you forget to define required blocks or use incorrect directives in the process definition. Maybe you try to use `inputs` rather than the correction `input` directive, as we illustrate in `invalid_process.nf`:
```
FriederikeHanssen: maybe start with: "Open the `invalida_process.nf". This feels a little backwards

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+    // Create input channel
+    input_ch = Channel.of('sample1', 'sample2', 'sample3')
+
+    // Call the process with the input channel
+    PROCESS_FILES(input_ch)
+}
+```
+
+Try to run this and you'll see an error like:
```
FriederikeHanssen: missing nextflow run command

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+       ^
+
+1 error
+
+
+ -- Check '.nextflow.log' file for details
+```
+
+In these cases the error messaging is quite straightforward, and you should be able to quickly replace the incorrect keyword with the correct on by referencing [the documentation](https://www.nextflow.io/docs/latest/process.html#).
+
```
FriederikeHanssen: I think we should add the solution for anyone doing self-taught training

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+```
+
+Try to run this and you'll see an error like:
+
+```console title="Invalid process keyword error"
+Nextflow 25.04.6 is available - Please consider updating your version to it
+
+ N E X T F L O W   ~  version 24.10.2
+
+Launching `invalid_process.nf` [happy_lorenz] DSL2 - revision: 2bfe3a8652
```
FriederikeHanssen: I did not get the error. Did you push the corrected file?

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+1 error
+
+
+ -- Check '.nextflow.log' file for details
+```
+
+In these cases the error messaging is quite straightforward, and you should be able to quickly replace the incorrect keyword with the correct on by referencing [the documentation](https://www.nextflow.io/docs/latest/process.html#).
+
+### 1.3. Using bad variable names
+
```
FriederikeHanssen: Next open the file `no_such_var.nf`

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    echo "Using undefined variable: ${undefined_var}" >> ${output_prefix}.txt  // ERROR: undefined_var not defined
+    """
+}
+
+workflow {
+    input_ch = Channel.of('sample1', 'sample2', 'sample3')
+    PROCESS_FILES(input_ch)
+}
+```
+
```
FriederikeHanssen, line 310: Maybe run first, see the error, then think about what is happening? This training feels like it could be very interactive with participants trying to find the error before telling them what it is

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+This will fail because `undefined_var` is not defined anywhere. The error message will indicate that the variable is not recognized in the script template. Run `no_such_var.nf` to see:
+
```
FriederikeHanssen, line 330: missing run command

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+  echo "Processing ${sample_name} on ${timestamp}" > ${output_prefix}.txt
+  echo "Using undefined variable: ${undefined_var}" >> ${output_prefix}.txt  // ERROR: undefined_var not defined
+  """
+
+Tip: when you have fixed the problem you can continue the execution adding the option `-resume` to the run command line
+
+ -- Check '.nextflow.log' file for details
+```
+
+If you get a 'No such variable' error it will be because you've made a mistake like this. You can fix it by either defining the variable (by correcting input variable names or editing groovy code before the script), or by removing it from the script block if it's not needed.
```
FriederikeHanssen: again I think we should add a solution for self-taught learners

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+```console title="No such variable error"
+ERROR ~ Error executing process > 'PROCESS_FILES (3)'
+
+Caused by:
+  No such variable: undefined_var -- Check script 'bad_var.nf' at line: 15
+
+
+Source block:
+  def output_prefix = "${sample_name}_processed"
+  def timestamp = new Date().format("yyyy-MM-dd")
```
FriederikeHanssen, line 340: I think this might be the first time we are showing that you can inject groovy code before the script block.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+Tip: when you have fixed the problem you can continue the execution adding the option `-resume` to the run command line
+
+ -- Check '.nextflow.log' file for details
+```
+
+If you get a 'No such variable' error it will be because you've made a mistake like this. You can fix it by either defining the variable (by correcting input variable names or editing groovy code before the script), or by removing it from the script block if it's not needed.
+
+### 1.4. Bad use of Bash variables
+
```
FriederikeHanssen, line 354: Open the file...

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    path "${sample_name}_output.txt"
+
+    script:
+    """
+    prefix="${sample_name}_output"
+    echo "Processing ${sample_name}" > ${prefix}.txt
+    """
+}
+```
+
```
FriederikeHanssen: Run the file and you should see..

```
  No such variable: prefix -- Check script 'bad_bash_var.nf' at line: 11
  ```

Comment (FriederikeHanssen) [reply docs/side_quests/debugging.md]: might be interesting as well because we are now getting error message in the more typical format that you would also normally see when running a pipeline and it dies for memory/cpu/tool failure reasons

File: docs/side_quests/debugging.md [hunk 19]
```diff
@@ -0,0 +1,1401 @@
...
+Channel structure errors are more subtle than syntax errors because the code is syntactically correct but the data shapes don't match what processes expect. These errors typically appear at runtime and require understanding of how data flows through your workflow.
+
+### 2.1. Wrong Number of Input Channels
+
+This error occurs when you pass a different number of channels than a process expects. Open `bad_number_inputs.nf` in VS Code to see an example:
+
+```groovy title="Wrong number of channels"
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
```
FriederikeHanssen:
```suggestion
```

not part of the file you are using

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+ N E X T F L O W   ~  version 24.10.2
+
+Launching `bad_number_inputs.nf` [fabulous_kirch] DSL2 - revision: 177eb7aa24
+
+Process `PROCESS_FILES` declares 1 input channel but 2 were specified
+
+ -- Check script 'bad_number_inputs.nf' at line: 22 or see '.nextflow.log' file for more details
+```
+
+To fix this you need to either combine the channels into tuples or pass only the required channel. The error message clearly states that the process expects 1 input channel but 2 were provided. For this specific example, the process expects a single channel and doesn't require the second channel, so we can fix it by passing only the `samples_ch` channel. More commonly, you might add additional inputs to a process and forget to update the workflow call accordingly, which can lead to this type of error. Fortunately this is one of the easier to understand and fix errors, as the error message is quite clear about the mismatch.
```
FriederikeHanssen: maybe we can make this a bit easier to digest by highlighting the line numbers in the code that are relevant

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+workflow {
+
+    reference_ch = Channel.of('baseline_reference')
+    input_ch = Channel.of('sample1', 'sample2', 'sample3')
+
+    PROCESS_FILES(reference_ch, input_ch)
+}
+```
+
+Here, we create two queue channels, a reference to be passed to all `PROCESS_FILE` instances, and a channel of sample names. The process expects both inputs, but we pass them as separate channels. When you run this workflow, it will execute without error, but it will only process the first sample:
```
FriederikeHanssen: have we talked about queue vs value channels in previous training sections?

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+```console title="Exhausted channel output"
```
FriederikeHanssen: missing run command

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+```
+
+This happens because the `reference_ch` channel is exhausted after the first process execution, despite the `input_ch` channel having more items to process.
+
+This pattern of a single invariable reference channel is very common, so this error crops up a lot. If you see a process only running once when you expect it to run multiple times, check that you're not exhausting a channel that should be reused.
+
+There are a couple of ways to address this. You can simply set the channel to be a value type:
+
+```groovy
+    reference_ch = Channel.value('baseline_reference')
```
FriederikeHanssen: also `collect()` is used a lot (maybe just in sarek)

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+That will allow a singleton channel to be reused again and again. In more complex scenarios, perhaps where you have a more comlex situation than a single reference channel, you can use the `combine` operator to create a new channel that combines the two channels into tuples:
+
+```groovy
+    reference_ch = Channel.of('baseline_reference')
+    input_ch = Channel.of('sample1', 'sample2', 'sample3')
+    combined_ch = reference_ch.combine(input_ch)
+```
+
+`.combine()` generates a cartesian product of the two channels, so each item in `reference_ch` will be paired with each item in `input_ch`. This allows the process to run for each sample while still using the reference.
```
FriederikeHanssen: maybe mention that this would require the process input to be adjusted as well and thus may not be a suitable solution in all cases.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    echo "Processing ${sample_name}" > ${sample_name}_output.txt
+    """
+}
+
+workflow {
+    PROCESS_FILES(input_ch)
+}
+```
+
+You may be able to see that we're generating a channel composed of tuples: `['sample1', 'file1.txt']`, but the process expects a single value, `val sample_name`. When you run this workflow:
```
FriederikeHanssen: I think highlighting the relevant lines could help spotting this

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    input_ch = Channel.of(
+        ['sample1', 'file1.txt'],
+        ['sample2', 'file2.txt']
+    )
+    .view { "Channel content: $it" }
+    .map { tuple -> tuple[0] }
+    .view { "After mapping: $it" }
+```
+
+With `.view()`, you can understand the shape of your channels at all stages to help with debugging. Run `bad_channel_shape_viewed.nf` to see this in action:
```
FriederikeHanssen: given the similar names, an explicit call out in the beginning would help I think

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+```groovy title="Extracting single value from tuple"
+    PROCESS_FILES(input_ch).map { it[0] }  // Extract first element
+```
+
+### 2.4. Channel Debugging Techniques
+
+#### Using `.view()` for Channel Inspection
+
+The most powerful debugging tool for channels is the `.view()` operator:
```
FriederikeHanssen: are we ever going to show `dump` 🥹 the best of all because you can leave it in your pipeline

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    """
+    echo "Processing ${sample_name}" > ${sample_name}_output.txt
+    """
+}
+```
+
+The error message indicates that the process expected to produce an output file named `sample3.txt`, but the script actually creates `sample3_output.txt`. This mismatch causes the process to fail.
+
+If you encounter this sort of error, go back and check that the outputs match between your process definition and your output block. If the problem still isn't clear, check the process directory itself to identify the actual output files created. You can do this by looking in the work directory for the process:
+
```
FriederikeHanssen, line 1138: I think a pointer to where this is set in the code and how to fix it would also be helpful here

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+```bash
+❯ ls -h work/02/9604d49fb8200a74d737c72a6c98ed
+sample3_output.txt
+```
+
+### 3.2. Missing software
+
+`missing_software.nf` is an example similar to the others we've encoutered in this module. There are no syntax errors, and this time we're using the `cowpy` command. If you try to run this workflow:
```
FriederikeHanssen:
```suggestion
`missing_software.nf` is an example similar to the others we've encountered in this module. There are no syntax errors, and this time we're using the `cowpy` command. If you try to run this workflow:
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+report {
+    enabled = true
+    file = "${params.output}/report.html"
+    overwrite = true
+}
+```
+
+We've done this for you in the `nextflow.config` file, so you can run most of the debugging examples in this guide and automatically get comprehensive debugging information.
+
+This generates three essential debugging files:
```
FriederikeHanssen: let's link out to the docs here, also because the trace file can be customized to give even more information if needed

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+### 4.2. Preview Mode and Process Debugging
+
+While trace files help you understand what went wrong after execution, sometimes you want to catch problems before any processes run, or see what's happening in real-time. Nextflow provides two key approaches for this kind of proactive debugging.
+
+#### Dry Run Testing
+
+The first approach is preview mode, which lets you test workflow logic without executing commands:
+
+```bash
+nextflow run workflow.nf -preview
```
FriederikeHanssen: this file does not exist

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    """
+
+    stub:
+    """
+    touch ${sample_name}_output.txt
+    """
+}
+```
+
+(you'll see this in `missing_software_stub.nf`)
```
FriederikeHanssen: would it make sense to just have people add the stub run into the existing file? Just to have something todo so to say

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+Once you've identified a problem using the techniques above, you need an efficient way to test your fixes without wasting time re-running successful parts of your workflow. This is where Nextflow's resume functionality becomes invaluable for debugging.
+
+#### Efficient Debug-Fix-Test Cycles
+
+The resume feature allows you to quickly iterate on fixes by only re-running failed or modified tasks:
+
+```bash
+# Initial run (fails at some point)
+nextflow run workflow.nf -with-trace
```
FriederikeHanssen: file does not exists

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+```
+
+**Investigate work directories** for failed tasks using the hash from trace:
+
+```bash
+# Use the actual hash from your trace file
+ls -la work/[hash]*/
+cat work/[hash]*/.command.err
+cat work/[hash]*/.command.sh
+```
```
FriederikeHanssen: depending on how one solves this, the error commands look way different. For example my first attempt i used processed_ch as input for heay_ch and ended with an error code that looks like this:

```
Caused by:
  File `/workspaces/training/side-quests/debugging/work/fa/9a752eb4288c0d1fbc0429d6e5a158/sample_002_result.txt_heavy.txt` is outside the scope of the process work directory: /workspaces/training/side-quests/debugging/work/5c/fe0505cdb6f69d2608c8c8b4a66d26


Command executed:

  # Simulate heavy computation
  for i in {1..1000000}; do
      echo "Heavy computation $i for /workspaces/training/side-quests/debugging/work/fa/9a752eb4288c0d1fbc0429d6e5a158/sample_002_result.txt"
  done > /workspaces/training/side-quests/debugging/work/fa/9a752eb4288c0d1fbc0429d6e5a158/sample_002_result.txt.txt

```

much harder to debug then the time one if you use `input_ch`

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+### Takeaway
+
+Effective Nextflow debugging combines understanding error messages, using built-in tools (trace files, work directories, preview mode), and applying systematic approaches. Practice with these examples builds the experience needed to debug real-world workflow issues efficiently.
+
+### What's next?
+
+Apply these debugging skills in your own workflow development and explore:
+
+- **[Development Best Practices](development_best_practices.md)**: Prevent bugs with systematic development approaches
```
FriederikeHanssen: pointing to own guide 
i think we should insist on the ide features to be run first

FriederikeHanssen (commented):
This ended up being a lot of comments, discard at your discretion. I think having some before/after, line highlighting, and more consistently adding in the "things" the user needs to do as commands (nextflow run commands, maybe updating the workflow files themselves instead of providing all of them) 

I envision this training to be quite interactive and hopefully allowing a lot of discussions with the participants trying to sort out the errors themselves

docs/side_quests/debugging.md
adamrtalbot: I'm not sure these bullet points are necessary. They're a bit verbose. Just saying familiar with x, y and z should be sufficient.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+This guide focuses on **systematic debugging techniques** for Nextflow workflows:
+
+- **Common error patterns**: Understanding the most frequent types of errors and how to identify them
+- **Syntax error debugging**: Using IDE features and Nextflow error messages effectively
+- **Channel debugging**: Diagnosing data flow issues and channel structure problems
+- **Process debugging**: Investigating execution failures and resource issues
+- **Built-in debugging tools**: Leveraging Nextflow's trace, timeline, and reporting features
+- **Advanced debugging techniques**: Using resume, stub running, and other advanced approaches
+- **Systematic debugging workflow**: Following a structured approach to problem-solving
```
adamrtalbot:
```suggestion
- **Common error patterns**: Understanding the most frequent types of errors and how to identify them
- **Syntax error debugging**: Using IDE features and Nextflow error messages effectively
- **Channel debugging**: Diagnosing data flow issues and channel structure problems
- **Process debugging**: Investigating execution failures and resource issues
- **Built-in debugging tools**: Leveraging Nextflow's trace, timeline, and reporting features
- **Advanced debugging techniques**: Using resume, stub running, and other advanced approaches
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+### 1.2. Using incorrect process keywords or directives
+
+Another common syntax error is an **invalid process definition**. This can happen if you forget to define required blocks or use incorrect directives in the process definition. Maybe you try to use `inputs` rather than the correction `input` directive, as we illustrate in `invalid_process.nf`:
+
+```groovy
+#!/usr/bin/env nextflow
+
+process PROCESS_FILES {
+    inputs:
```
adamrtalbot: when I opened `invalid_process.nf` it actually had `input:` here, so it worked fine.

docs/side_quests/debugging.md
adamrtalbot: Agreed!

docs/side_quests/debugging.md
adamrtalbot: Agreed, I would add a `nextflow run` here. It's what I immediately did.

docs/side_quests/debugging.md
adamrtalbot: In this case, we need to know which file because it isn't clear.

adamrtalbot: (it's `bad_bash_var.nf`)

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+}
+```
+
+Starting out in Nextflow, it can be difficult to understand the difference between Nextflow (Groovy) and Bash variables. In this example, we're defining the `prefix` variable in Bash, but the syntax we used to refer to it (`${prefix}`) is Groovy syntax, not Bash syntax. The variable doesn't exist in the Groovy context, so we will get a 'no such variable' error if we try to run the workflow (you can do this by running `bad_bash_var.nf`).
+
+If you want to use a Bash variable, you must escape the dollar sign like this:
+
+```groovy title="Using Bash variables correctly"
+echo "Processing ${sample_name}" > \${prefix}.txt
+```
```
adamrtalbot: It might help to indicate why this happens. 

I would replace the tip below with what is actually occurring here; the script stuff is a bit irrelevant.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    def output_prefix = "${sample_name}_processed"
+    def output_file = "${output_prefix}.txt"
+    """
+    echo "Processing ${sample_name}" > ${output_file}
+    """
+    ```
+
+    This approach avoids the need to escape dollar signs and makes the code easier to read and maintain.
+
+### 1.5. Non-lethal syntax errors when using the VSCode extension
```
adamrtalbot: I'm not sure I will call it non-lethal.

Bad practice? Poor standards?

More importantly, add _why_ it's not a good idea (encapsulation of the logic, uncharacterized variables from global scope).

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+When you run this workflow, it will execute successfully, but the VSCode extension will highlight the `input_ch` variable as being defined outside the workflow block, which is not recommended:
+
+![Non-lethal syntax error](img/nonlethal.png)
+
+This is a non-lethal syntax error that won't prevent execution but could lead to confusion or unexpected behavior in larger workflows. Tighter restrictions on such things will likely become enforced in future Nextflow versions, so for this example it's good practice to keep your input channels defined within the workflow block, and in general to follow any other recommendations the extension makes.
+
+### Takeaway
+
+You can systematically identify and fix syntax errors using Nextflow error messages and IDE visual indicators. Common syntax errors include missing braces, incorrect process keywords, undefined variables, and improper use of Bash vs. Nextflow variables. The VSCode extension helps catch many of these before runtime.
```
adamrtalbot: This is a bit meh, if you can give it a stronger ending (tell 'em what you told 'em) it will help round off the section.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+### What's next?
+
+Learn to debug more complex channel structure errors that occur even when syntax is correct.
+
+---
+
+## 2. Channel Structure Errors
+
+Channel structure errors are more subtle than syntax errors because the code is syntactically correct but the data shapes don't match what processes expect. These errors typically appear at runtime and require understanding of how data flows through your workflow.
```
adamrtalbot:
```suggestion
Channel structure errors are more subtle than syntax errors because the code is syntactically correct, but the data shapes don't match what processes expect. Nextflow will try to run the pipeline, but might find that the number of files doesn't match what it expects and fail. These errors typically only appear at runtime and require an understanding of the data flowing through your workflow.
```

adamrtalbot: A little refinement to the paragraph.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+This error occurs when you pass a different number of channels than a process expects. Open `bad_number_inputs.nf` in VS Code to see an example:
+
+```groovy title="Wrong number of channels"
+#!/usr/bin/env nextflow
+nextflow.enable.dsl=2
+
+// Create two separate channels
+samples_ch = Channel.of('sample1', 'sample2', 'sample3')
+files_ch = Channel.of('file1.txt', 'file2.txt', 'file3.txt')
```
adamrtalbot: Why is this bit outside the workflow scope?

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+```console title="Wrong number of channels error"
+ N E X T F L O W   ~  version 24.10.2
+
+Launching `bad_number_inputs.nf` [fabulous_kirch] DSL2 - revision: 177eb7aa24
+
+Process `PROCESS_FILES` declares 1 input channel but 2 were specified
+
+ -- Check script 'bad_number_inputs.nf' at line: 22 or see '.nextflow.log' file for more details
+```
+
```
adamrtalbot: State the error again here, in human not Nextflow.

docs/side_quests/debugging.md
adamrtalbot: Talking about combining channels into tuples is confusing, so let's just trim off that bit:
```suggestion
The error message clearly states that the process expects 1 input channel, but 2 were provided. For this specific example, the process expects a single channel and doesn't require the second channel, so we can fix it by passing only the `samples_ch` channel. More commonly, you might add additional inputs to a process and forget to update the workflow call accordingly, which can lead to this type of error. Fortunately, this is one of the easier-to-understand and fix errors, as the error message is quite clear about the mismatch.
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+To fix this you need to either combine the channels into tuples or pass only the required channel. The error message clearly states that the process expects 1 input channel but 2 were provided. For this specific example, the process expects a single channel and doesn't require the second channel, so we can fix it by passing only the `samples_ch` channel. More commonly, you might add additional inputs to a process and forget to update the workflow call accordingly, which can lead to this type of error. Fortunately this is one of the easier to understand and fix errors, as the error message is quite clear about the mismatch.
+
+### 2.2. Mismatched channel arity
+
+Some channel structure errors are much more subtle and produce no errors at all. Probably the most common of these reflects a challenge that new Nextflow users face in understanding that channels can be exhausted. Open `exhausted.nf` in VS Code to see an example:
```
adamrtalbot:
```suggestion
Some channel structure errors are much more subtle and produce no errors at all. Probably the most common of these reflects a challenge that new Nextflow users face in understanding that queue channels can be exhausted and run out of samples, meaning the workflow finishes prematurely. Open `exhausted.nf` in VS Code to see an example:
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+```console title="Exhausted channel output"
+ N E X T F L O W   ~  version 24.10.2
+
+Launching `exhausted.nf` [magical_mayer] DSL2 - revision: 4ec9570696
+
+executor >  local (1)
+[b3/42dbdb] process > PROCESS_FILES (1) [100%] 1 of 1 ✔
+```
+
+This happens because the `reference_ch` channel is exhausted after the first process execution, despite the `input_ch` channel having more items to process.
```
adamrtalbot: I'm tempted to:
1. Run the workflow
2. See the error
3. Check the code
4. See why the error has happened

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+This pattern of a single invariable reference channel is very common, so this error crops up a lot. If you see a process only running once when you expect it to run multiple times, check that you're not exhausting a channel that should be reused.
```
adamrtalbot:
```suggestion
This pattern of a single reference file in a channel is very common, so this error occurs frequently. If you see a process only running once when you expect it to run multiple times, check that you're not exhausting a channel that should be reused.
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+There are a couple of ways to address this. You can simply set the channel to be a value type:
```
adamrtalbot:
```suggestion
There are a couple of ways to address this. You can simply create a value channel type, which can be used over and over again:
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+```groovy
+    reference_ch = Channel.value('baseline_reference')
+```
+
+That will allow a singleton channel to be reused again and again. In more complex scenarios, perhaps where you have a more comlex situation than a single reference channel, you can use the `combine` operator to create a new channel that combines the two channels into tuples:
```
adamrtalbot: Removing the initial sentence and moving it above to keep the paragraph saying just 1 thing.

```suggestion
In more complex scenarios, perhaps where you have a more complex situation than a single reference channel, you can use the `combine` operator to create a new channel that combines the two channels into tuples:
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    reference_ch = Channel.of('baseline_reference')
+    input_ch = Channel.of('sample1', 'sample2', 'sample3')
+    combined_ch = reference_ch.combine(input_ch)
+```
+
+`.combine()` generates a cartesian product of the two channels, so each item in `reference_ch` will be paired with each item in `input_ch`. This allows the process to run for each sample while still using the reference.
+
+### 2.3. Wrong Channel Content Structure
+
+When workflows reach a certain level of complexity, it can be a little difficult to keep track of the internal structures of each channel, and people commonly generate mismatches between what the process expects and what the channel actually contains. This is different to the issue we discussed earlier where the number of channels was incorrect, and is more subtle. The channel may have the right number of items, but the structure of those items doesn't match what the process expects.
```
adamrtalbot: Some language stuff

```suggestion
When workflows reach a certain level of complexity, it can be a little difficult to keep track of the internal structures of each channel, and people commonly generate mismatches between what the process expects and what the channel actually contains. This is more subtle to the issue we discussed earlier, where the number of channels was incorrect. In this case, the channel may have the right number of items, but the structure of those items doesn't match what the process expects.
```

docs/side_quests/debugging.md
adamrtalbot: This could be expanded a bit. How about providing two solutions, for different scenarios:
1. 1 reference file used a lot: `Channel.value`, `.first()` or `.collect()`
2.  multiple files used for all samples in the sample channel `.combine()`

You could do a complete example for both. As a first pass, you could just do (1) and note (2).

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+```groovy title="Fixing channel structure"
+    tuple val(sample_name), path(file_name)
+```
+
+Or if, as in this example the process only needs the sample name, we can extract the first element of the tuple before passing it to the process:
+
+```groovy title="Extracting single value from tuple"
+    PROCESS_FILES(input_ch).map { it[0] }  // Extract first element
+```
```
adamrtalbot: I'd add a before/after and a successful run here (and for most examples). We could use the hidden feature used in the nextflow run module to keep it concise.

docs/side_quests/debugging.md
adamrtalbot: 💩

docs/side_quests/debugging.md
adamrtalbot: Rather than having separate files, I'm tempted to chain these into a before/after and build it in iterations.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+}
+```
+
+So all we need to to do is run the workflow with Docker enabled. We've set up a Docker profile for you in `nextflow.config`, so you can run the workflow with:
+
+```bash
+nextflow run missing_software.nf -profile docker
+```
+
+... and everything should run successfully.
```
adamrtalbot: Let's add a note here pointing to "hello containers"

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    val sample_name
+
+    output:
+    path "${sample_name}_output.txt"
+
+    script:
+    """
+    cowpy ${sample_name} > ${sample_name}_output.txt
+    """
+}
```
adamrtalbot: How about using sleep to force it?

```suggestion
process PROCESS_FILES {

    time '1 ms'

    input:
    val sample_name

    output:
    path "${sample_name}_output.txt"

    script:
    """
    sleep 1
    cowpy ${sample_name} > ${sample_name}_output.txt
    """
}
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+nextflow run bad_resources.nf -profile docker
+```
+
+... gives us an error like:
+ERROR ~ Error executing process > 'PROCESS_FILES (1)'
+
+Caused by:
+Process exceeded running time limit (1ms)
+
+````
```
adamrtalbot:
```suggestion
... gives us an error:

\```console
ERROR ~ Error executing process > 'PROCESS_FILES (1)'

Caused by:
Process exceeded running time limit (1ms)

\```
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+Nextflow provides several powerful built-in tools for debugging and analyzing workflow execution. These tools help you understand what went wrong, where it went wrong, and how to fix it efficiently.
+
+### 4.1. Execution Reports and Trace Files
+
+#### Generate Complete Debugging Information
+
+Always run with comprehensive reporting when debugging, either by adding flags to your command line:
+
+```bash
+nextflow run workflow.nf -with-trace -with-timeline -with-report
```
adamrtalbot: Is this really debugging? This is more like metrics and performance to me. I would remove it

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+```bash
+grep "FAILED" results/trace.txt
+```
+
+#### Work Directory Investigation
+
+In [Hello Nextflow](../hello_nextflow/) you learned that each task creates a work directory. This directory contains all the files generated during execution, including the command script, output files, and logs. You can find the work directory for a failed task using its hash from the trace file.
+
+Each task creates a work directory with complete debugging information:
```
adamrtalbot: Definitely drop the tracing stuff (or move it right to the end). It doesn't add much and this is more focused on debugging.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+```bash
+# Use the hash from trace.txt to find the work directory
+ls -a work/1c/c77799*/
```
adamrtalbot: Let's list the contents nicely:

```suggestion
ls -1a work/1c/c77799*/
```

docs/side_quests/debugging.md
adamrtalbot: I use `ls -1a` to list all files downwards without file sizes or permissions.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+- **`.command.sh`**: The exact script that was executed
+- **`.command.run`**: The command wrapper used to run the task. This includes environment variables and other settings, for example the container image used.
+- **`.command.err`**: Standard error output (most important for debugging)
+- **`.command.out`**: Standard output from the command
+- **`.command.log`**: Nextflow's execution log
+
+### 4.2. Preview Mode and Process Debugging
+
+While trace files help you understand what went wrong after execution, sometimes you want to catch problems before any processes run, or see what's happening in real-time. Nextflow provides two key approaches for this kind of proactive debugging.
```
adamrtalbot: You don't actually use the trace files for debugging, so perhaps keep it focused on the `.command` files

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+The first approach is preview mode, which lets you test workflow logic without executing commands:
+
+```bash
+nextflow run workflow.nf -preview
+```
+
+This can be quite useful for quickly checking the structure of your workflow and ensuring that processes are connected correctly without running any actual commands. For example, for our first syntax errror from earlier:
+
+```bash
+nextflow run bad_syntax.nf -preview
```
adamrtalbot: if they've fixed this, it wont work!

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+ -- Check '.nextflow.log' file for details
+```
+
+#### Real-time Process Output
+
+Preview mode catches structural issues before execution, but sometimes you need to see what's happening inside running processes. The second approach is enabling real-time process output, which shows you exactly what each task is doing as it executes.
+
+For example, `bad_channel_shape_viewed.nf` from our earlier examples printed channel content using `.view()`. We can also use the `debug` directive to echo variables from within the process itself, which we demonstrate in `bad_channel_shape_viewed_debug.nf`:
+
```
adamrtalbot: We could incorporate this into the channel shape bit so it's all 1 flow?

adamrtalbot: This also sticks out because it's not a cli argument like `-preview` or `-stub`, so moving it up there will make it more consistent.

docs/side_quests/debugging.md
adamrtalbot: yep, nothing wrong with not spoon feeding someone the answer. We can add a `solutions/` dir for the solutions.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+workflow {
+    // Minimal test input
+    test_input = Channel.of(['test_sample', file('data/small_test.fastq')])
+
+    PROBLEMATIC_PROCESS(test_input)
+
+    // Inspect output
+    PROBLEMATIC_PROCESS.out.view { "Process output: $it" }
+}
+```
```
adamrtalbot: I don't like this bit! I think it can be removed.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+Not all workflow failures are due to syntax or logic errors. In production environments, many debugging challenges stem from resource constraints - processes that run out of memory, exceed time limits, or compete for system resources. Understanding how to diagnose and fix these issues is crucial for reliable workflow execution.
+
+#### Identifying Resource Issues
+
+Resource problems often manifest as specific exit codes that give you immediate clues about what went wrong:
+
+- **Exit 130**: Process killed (often CTRL+C)
+- **Exit 137**: Process killed by system (usually out of memory)
+- **Exit 127**: Command not found
```
adamrtalbot: One thing to note here is Nextflow directly reports the error from the host system.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+    errorStrategy { task.exitStatus in [130,140] ? 'retry' : 'finish' }
+    maxRetries 2
+
+    script:
+    """
+    echo "Attempt ${task.attempt} with ${task.memory} memory"
+    your_command --memory ${task.memory.toGiga()}G
+    """
+}
+```
```
adamrtalbot: Does this...say anything? I feel like we need to get a message across.

I think the lesson should:

> #### Errors from the process itself
>
> Nextflow is deliberately simple - it will submit tasks to the host machine or scheduler, which are executed there and transparently show any errors. Things go wrong within tools, syntax errors or invalid files. When a Nextflow pipeline fails, take a moment to check the error and see if it originates from _Nextlow_ or from the _tool_.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+**Phase 3: Fix and Validate (15 minutes)**
+
+1. Make minimal targeted fixes
+2. Test with resume: `nextflow run workflow.nf -resume`
+3. Verify complete workflow execution
+
+#### Debugging Configuration Profile
+
+To make this systematic approach even more efficient, you can create a dedicated debugging configuration that automatically enables all the tools you need:
```
adamrtalbot, line 1729: Nah, the lesson has been about debugging and strategy. No need to switch to configuration now.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+
+Use the debug profile:
+
+```bash
+nextflow run workflow.nf -profile debug
+```
+
+### 4.7. Practical Debugging Exercise
+
+Now it's time to put the systematic debugging approach into practice. The workflow `buggy_workflow.nf` contains several common errors that represent the types of issues you'll encounter in real-world development. This exercise will help you experience how the three-phase debugging method works with actual problems.
```
adamrtalbot: I like this idea. Rob's old material has a hidden "answers" section we can use for each problem.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1401 @@
...
+### What's next?
+
+Apply these debugging skills in your own workflow development. The more you practice these techniques, the faster and more effective you'll become at identifying and resolving issues. Consider:
+
+- **Building robust workflows** with built-in debugging from the start
+- **Contributing to the community** by sharing debugging solutions
+- **Developing testing strategies** to catch issues before they reach production
+- **Mentoring others** using the systematic approaches you've learned
+
+Remember: effective debugging is a skill that improves with practice. The systematic methodology and comprehensive toolkit you've developed here will serve you well throughout your Nextflow development journey.
```
adamrtalbot: testing with nf-test?

adamrtalbot (commented):
I like the content, but I agree with Rike, every section needs a consistent syntax. 

1. Run the pipeline
2. Check the code
3. Fix the code
4. Run the pipeline

- Some bits can be iterative (the bad_channel stuff).
- The trace stuff can go, I don't think it adds much. We should "replace" it with a section on how some errors are from the host and Nextflow is behaving properly.

docs/side_quests/debugging.md
FriederikeHanssen: 3e9e0c1a

docs/side_quests/debugging.md
FriederikeHanssen: c433dab1

docs/side_quests/debugging.md
FriederikeHanssen: fixed in 737d2670

docs/side_quests/debugging.md
FriederikeHanssen: 647c9ad7

docs/side_quests/debugging.md
FriederikeHanssen: 647c9ad7

docs/side_quests/debugging.md
FriederikeHanssen: b986d865

docs/side_quests/debugging.md
FriederikeHanssen: b986d865

docs/side_quests/debugging.md
FriederikeHanssen: a660a462

docs/side_quests/debugging.md
FriederikeHanssen: d0377ae6

docs/side_quests/debugging.md
FriederikeHanssen: b1740977

docs/side_quests/debugging.md
FriederikeHanssen: reshuffled in b1740977

docs/side_quests/debugging.md
FriederikeHanssen: added in b1740977

docs/side_quests/debugging.md
FriederikeHanssen: 1bc7c3a4 addressed a little here, can use another pass

docs/side_quests/debugging.md
FriederikeHanssen: a0669683

docs/side_quests/debugging.md
FriederikeHanssen: 39706ecb

docs/side_quests/debugging.md
FriederikeHanssen: what do you mean?

FriederikeHanssen: added a note to reintroduce the error if they fixed it.

FriederikeHanssen: 9a5640f2

docs/side_quests/debugging.md
FriederikeHanssen: reshuffeled a bit and removed trace 026fbed9

docs/side_quests/debugging.md
FriederikeHanssen: moved the debug up into it's own section

docs/side_quests/debugging.md
FriederikeHanssen: done in 026fbed9

docs/side_quests/debugging.md
FriederikeHanssen: done c589a07c

docs/side_quests/debugging.md
FriederikeHanssen: ed7fbb28

docs/side_quests/debugging.md
FriederikeHanssen: I removed some of the tracing stuff because of the comment above and added your suggestion. I still find this section a bit weak. We probably need to add in a stronger example

docs/side_quests/debugging.md
FriederikeHanssen: simplified it but leaving it for time reasons right now

docs/side_quests/debugging.md
FriederikeHanssen: expanded point above

FriederikeHanssen: 5ee74d31

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,1525 @@
...
+Effective Nextflow debugging combines understanding error messages, using built-in tools (trace files, work directories, preview mode), and applying systematic approaches. Practice with these examples builds the experience needed to debug real-world workflow issues efficiently.
+
+### What's next?
+
+Apply these debugging skills in your own workflow development and explore:
+
+- **[Development Best Practices](development_best_practices.md)**: Prevent bugs with systematic development approaches
+- **[IDE Features](ide_features.md)**: Leverage VS Code features for faster error detection
+
+---
```
FriederikeHanssen:
```suggestion
---
```

think we can remove this. The development guide is cancelled for now. We are pointing to the IDE one in the beginning

docs/side_quests/debugging.md
FriederikeHanssen: discussed in person. we'll think about it later

docs/side_quests/debugging.md
FriederikeHanssen: removed the section on tracing

docs/side_quests/debugging.md
FriederikeHanssen: renamed to bad-practice

docs/side_quests/debugging.md
FriederikeHanssen: removed

docs/side_quests/debugging.md
FriederikeHanssen: removed

docs/side_quests/debugging.md
FriederikeHanssen: removed

docs/side_quests/debugging.md
pinin4fjords: OK, copied that approach, and tidied up the exercise a bit

docs/side_quests/debugging.md
pinin4fjords: Agreed. We cover it in Hello Genomics, but we should probably slide it into Hello Nextflow at some point. I don't think this is the material to teach it in explicitly.

docs/side_quests/debugging.md
pinin4fjords: If it's not somewhere in Hello Nexflow it definitely should be.

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,2093 @@
...
+```groovy title="exhausted.nf (fixed - Option 1a)" hl_lines="2" linenums="21"
+workflow {
+    reference_ch = Channel.value('baseline_reference')  // Value channel can be reused
+    input_ch = Channel.of('sample1', 'sample2', 'sample3')
+
+    PROCESS_FILES(reference_ch, input_ch)
+}
+```
+
+2. Use the `first()` [operator](https://www.nextflow.io/docs/latest/reference/operator.html#first):
```
FriederikeHanssen: the numbering in the rendered version :shake_fist: 
 
Screenshot 2025-08-05 at 10 43 35

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,2093 @@
...
+Tip: when you have fixed the problem you can continue the execution adding the option `-resume` to the run command line
+
+ -- Check '.nextflow.log' file for details
+```
+
+#### Check the code
+
+The error message indicates that the process expected to produce an output file named `sample3.txt`, but the script actually creates `sample3_output.txt`. Let's examine the process definition in `missing_output.nf`:
+
+```groovy title="missing_output.nf" linenums="3" hl_lines="2 9"
```
FriederikeHanssen:
```suggestion
```groovy title="missing_output.nf" linenums="3" hl_lines="6 10"
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,2093 @@
...
+
+    Use the systematic debugging approach to identify and fix all errors in `buggy_workflow.nf`. This workflow attempts to process sample data from a CSV file but contains multiple intentional bugs representing common debugging scenarios.
+
+    Start by running the workflow to see the first error:
+
+    ```bash
+    nextflow run buggy_workflow.nf
+    ```
+
+    Apply the four-phase debugging method you've learned:
```
FriederikeHanssen, line 1767:
```suggestion
    Apply the four-phase debugging method you've learned:
    
```

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,2093 @@
...
+
+Effective Nextflow debugging combines understanding error messages, using built-in tools (trace files, work directories, preview mode), and applying systematic approaches. Practice with these examples builds the experience needed to debug real-world workflow issues efficiently.
+
+---
+
+## Summary
+
+In this side quest, we've learned:
+
+1. **How to identify and fix syntax errors**:
```
FriederikeHanssen: numbering :screams: 
Screenshot 2025-08-05 at 10 56 13

docs/side_quests/debugging.md
```diff
@@ -0,0 +1,2094 @@
+# Debugging Nextflow Workflows
```
FriederikeHanssen:
```suggestion
# Debugging Workflows
```

FriederikeHanssen (approved):

docs/side_quests/debugging.md
FriederikeHanssen: how knew making numbers bold fixes this

adamrtalbot: If you indent the code and text between the bullets by 4 spaces, it considers it within the bullet point and fixes it as well.

FriederikeHanssen: hm prettier keeps moving it to three spaces.

FriederikeHanssen: https://github.com/nextflow-io/training/pull/639/commits/75089b3649731838a4c2ebccca539ae85d6d23d9 doesn't work. prettier insists on three spaces and then it is back to being rendered as 1.,1.,1., same in the summary. 
I am not going to spend much time on this now. Will just revert to workaround from before.

docs/side_quests/debugging.md
pinin4fjords: Let's leave it this way for now- it's robust, and it allows us to be consistent with the pattern we use elsewhere.

PR #641, opened 2025-08-01
Update advanced training to current syntax

adamrtalbot:
Updates to the advanced training to use the current syntax making it more consistent with the other guides.

Changes:
 - Replace all pipe operators with method chaining
 - Where required, assigns to intermediate channel and uses in a process
 - Moves all logic into workflows
 - Uses named variables in closures
 - Removes references to _not_ using implicit variable in closure (`it`)
 - Add `script:` block to every process
 - Removes supplying a multi-channel object to a process in favour of supplying each sub-channel individually to avoid confusion.

### DIFF
--- a/docs/advanced/structure.md
+++ b/docs/advanced/structure.md
@@ -58,7 +58,7 @@ workflow {
 
 To do this, we can create the bin directory, write our R script into the directory. Finally, and crucially, we make the script executable. This is the code we used to create the `cars.R` script, no need to run it:
 
-```bash linenums="1"
+```bash
 mkdir -p bin
 cat << EOF > bin/cars.R
 #!/usr/bin/env Rscript
@@ -89,7 +89,7 @@ code work/*/*/.command.run
 
 You'll notice a `nxf_container_env` bash function that appends our bin directory to `$PATH`:
 
-```bash
+```bash linenums="14"
 nxf_container_env() {
 cat << EOF
 export PATH="\$PATH:/workspaces/training/nf-training-advanced/structure/bin"
@@ -113,7 +113,7 @@ If a process script block is becoming too long, it can be moved to a template fi
 
 The structure directory already contains an example template - a very simple python script. We can add a new process that uses this template:
 
-```groovy linenums="1"
+```groovy linenums="14"
 process SayHiTemplate {
     debug true
     input: val(name)
@@ -156,7 +156,7 @@ workflow {
 
 We can use the new `hi` method in the workflow:
 
-```groovy linenums="1"
+```groovy linenums="1" hl_lines="3"
 workflow {
     Channel.of("Montreal")
         .map { new Metadata() }
@@ -166,7 +166,7 @@ workflow {
 
 At the moment, the `Metadata` class is not making use of the "Montreal" being passed into the closure. Let's change that by adding a constructor to the class:
 
-```groovy linenums="1"
+```groovy linenums="1" hl_lines="3 7"
 class Metadata extends HashMap {
     Metadata(String location) {
         this.location = location
@@ -180,7 +180,7 @@ class Metadata extends HashMap {
 
 Which we can use like so:
 
-```groovy linenums="1"
+```groovy linenums="1" hl_lines="3"
 workflow {
     Channel.of("Montreal")
         .map { place -> new Metadata(place) }
@@ -208,7 +208,7 @@ workflow {
 
 Why might this be helpful? You can add extra classes to the metadata which can be computed from the existing metadata. For example, we might want to add a method to get the adapter prefix into our Metadata class:
 
-```groovy
+```groovy linenums="10"
 def getAdapterStart() {
     this.adapter?.substring(0, 3)
 }
@@ -216,7 +216,7 @@ def getAdapterStart() {
 
 Which we might use like so:
 
-```groovy linenums="1"
+```groovy linenums="1" hl_lines="4 10"
 process UseMeta {
     input: val(meta)
     output: path("out.txt")
@@ -235,7 +235,7 @@ workflow {
 
 You might even want to reach out to external services such as a LIMS or the E-utilities API. Here we add a dummy "getSampleName()" method that reaches out to a public API:
 
-```groovy
+```groovy linenums="13"
 def getSampleName() {
     def get = new URL('https://postman-echo.com/get?sampleName=Fido').openConnection()
     def getRC = get.getResponseCode();
@@ -276,7 +276,7 @@ process UseMeta {
 
         We might increase the length of the adapter prefix to 5 characters:
 
-        ```groovy linenums="1"
+        ```groovy linenums="12"
             def getAdapterStart() {
                 this.adapter?.substring(0, 5)
             }
@@ -295,7 +295,7 @@ class Dog {
 
 We can create a new dog at the beginning of the workflow:
 
-```groovy linenums="1"
+```groovy linenums="1" hl_lines="3"
 workflow {
     def dog = new Dog(name: "fido")
     log.info "Found a new dog: ${dog.name}"
@@ -304,7 +304,7 @@ workflow {
 
 We can pass objects of our class through channels. Here we take a channel of dog names and create a channel of dogs:
 
-```groovy linenums="1"
+```groovy linenums="1" hl_lines="2-4"
 workflow {
     Channel.of("Argente", "Absolon", "Chowne")
         .map { name -> new Dog(name: name) }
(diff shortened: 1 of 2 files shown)


docs/advanced/operators.md
```diff
@@ -23,23 +23,17 @@ cd operators
...
-    Channel.of( 1, 2, 3, 4, 5 )
-    | map { num -> num * num }
-    | view
-}
-```
+The value passed into `map` is assigned to a variable using the `->` syntax (e.g., `num` in `{ num -> ... }`). This variable is local to the closure and used for any transformations inside it.
 
 Groovy is an optionally typed language, and it is possible to specify the type of the argument passed to the closure.
 
 ```groovy linenums="1"
```
FriederikeHanssen:
```suggestion
```groovy linenums="1" hl_lines="3"
```

docs/advanced/operators.md
```diff
@@ -600,9 +577,9 @@ We can supply arguments `name` and `newLine` to the `collectFile` operator to re
 
 ```groovy
```
FriederikeHanssen:
```suggestion
```groovy hl_lines="3"
```

docs/advanced/operators.md
```diff
@@ -623,9 +600,9 @@ By default, the collected file is written into the work directory, which makes i
...
```
FriederikeHanssen:
```suggestion
```groovy hl_lines="3"
```

docs/advanced/grouping.md
```diff
@@ -212,20 +238,23 @@ process CombineBams {
...
 
 ```groovy linenums="1"
-MapReads( samples, reference )
-| map { meta, bam ->
-    key = groupKey(meta.subMap('id', 'type'), meta.repeatcount)
-    [key, bam]
-}
-| groupTuple
-| CombineBams
+mapped_reads = MapReads( samples, reference )
```
FriederikeHanssen, line 248: the line number is starting to fall apart here. Maybe we just remove it? (Not a super important issue if we want to defer this to later)

docs/advanced/grouping.md
```diff
@@ -287,43 +312,49 @@ process MergeGenotyped {
...
+        [key, bam]
+    }
+    .groupTuple()
+
+combined_bams = CombineBams(mapped_reads)
+    .map { meta, bam -> [meta.subMap('id', 'type'), bam] }
+    .combine( intervals )
+
+genotyped_bams = GenotypeOnInterval(combined_bams)
+    .map { groupKey, bamfile -> [groupKey.target, bamfile] }
```
FriederikeHanssen: this is returning null in the meta map:

before `.map`:
```
[id:sampleA, type:normal], 
```

after .map:

```
[null, ...bam
```

docs/advanced/structure.md
FriederikeHanssen, line 1: for the ./lib example, not sure if it is worth the hassle, but nf-core/rnaseq has moved away from the lib dir a long time ago. We could mention the version and fix the link to point to the last version supporting it. the current link lands at a 404

docs/advanced/structure.md
```diff
@@ -152,8 +152,8 @@ We can then use this class in our workflow:
 ```groovy linenums="1"
 workflow {
     Channel.of("Montreal")
-    | map { new Metadata() }
-    | view
+        .map { new Metadata() }
```
FriederikeHanssen, line 152: the language server is upset about this:

`Metadata` is not definednextflow

it still runs

docs/advanced/structure.md
```diff
@@ -226,11 +227,12 @@ process UseMeta {
...
-    | UseMeta
-    | view
+    place_ch = Channel.of("Montreal")
+        .map { place -> new Metadata(place) }
+        .map { metadata -> metadata + [adapter:"AACGTAGCTTGAC"] }
+
+    UseMeta(place_ch)
+        .view()
 }
 ```
```
FriederikeHanssen, line 235: I am not sure what i am doing wrong with the getSampleName function but it is not compiling for me. Might be me

docs/advanced/structure.md
```diff
@@ -292,7 +294,7 @@ We can create a new dog at the beginning of the workflow:
 
 ```groovy linenums="1"
 workflow {
-    dog = new Dog(name: "fido")
+    def dog = new Dog(name: "fido")
```
FriederikeHanssen, line 300: same here, language server doesn't like lib classes

docs/advanced/structure.md
adamrtalbot: Let's remove the reference to nf-core/rnaseq.

FriederikeHanssen (commented):
- the training "Configuration" needs a`docker.enabled=true` added to the nextflow.config

docs/advanced/grouping.md
adamrtalbot: `.target` is now a private attribute: https://github.com/nextflow-io/nextflow/blob/cfd0f306cb16745dc82d179afc7f6923234c3486/modules/nextflow/src/main/groovy/nextflow/extension/GroupKey.groovy#L35

adamrtalbot: We can use `as Map` instead.

docs/advanced/structure.md
adamrtalbot: Removed this stuff

docs/advanced/structure.md
adamrtalbot: Yup, I'll add a note.

docs/advanced/structure.md
adamrtalbot: It needs to include jsonslurper:

```
/Users/adam.talbot/training/nf-training-advanced/structure/lib/Metadata.groovy: 18: unable to resolve class JsonSlurper
 @ line 18, column 25.
               JsonSlurper jsonSlurper = new JsonSlurper()
                           ^

/Users/adam.talbot/training/nf-training-advanced/structure/lib/Metadata.groovy: 18: unable to resolve class JsonSlurper
 @ line 18, column 39.
               JsonSlurper jsonSlurper = new JsonSlurper()
```

adamrtalbot: Fixed

docs/advanced/structure.md
adamrtalbot: Left a note about it but it'll have to do. This is rare stuff anyway.

docs/advanced/grouping.md
adamrtalbot: Removed the line numbers but added lots of highlighting to make it more clear.

adamrtalbot: I've gone through and corrected all the line numbers and highlights now.

FriederikeHanssen:
you can ignore me, I scrolled passed these instructions https://training.nextflow.io/latest/advanced/configuration/#system-wide-configuration-homenextflowconfig and just tried running it directly

FriederikeHanssen (approved):

docs/advanced/metadata.md
```diff
@@ -133,68 +133,68 @@ We are almost there, but we still don't have the "treatment" metadata captured i
...
+            def (sample, replicate, type) = id.tokenize("_")
+            def (treatmentFwd, treatmentRev) = reads*.parent*.name*.minus(~/treatment/)
+            def meta = [
+                sample:sample,
+                replicate:replicate,
+                type:type,
+                treatmentFwd:treatmentFwd,
+                treatmentRev:treatmentRev,
+            ]
+            [meta, reads]
```
bentsherman, line 189: The future type checker will like it if you create tuples with the `tuple()` function instead of a list

docs/advanced/structure.md
```diff
@@ -248,6 +247,12 @@ def getSampleName() {
 }
 ```
 
+This relies on jsonSlurper which isn't included by default. Import this by adding the following to the top of the file:
+
+```groovy linenums="1"
+import groovy.json.JsonSlurper
+```
```
bentsherman, line 254: You can also just use JsonSlurper without importing it as `new groovy.json.JsonSlurper()`

bentsherman (commented):
I'm late to the party but just wanted to add a few suggestions.

Also note that channel factories should be accessed via the `channel` namespace instead of the `Channel` type. Should be a simple search-and-replace :smile:

docs/advanced/grouping.md
bentsherman: FWIW you can still use `.getGroupTarget()` as shown in the [docs](https://nextflow.io/docs/latest/reference/operator.html#grouptuple)

docs/advanced/structure.md
bentsherman: I think the issue is that the language server only checks for `lib` in the root directory of the workspace. So if you have pipeline examples in nested directories, the nested `lib` directory won't be used.

Unfortunately I haven't figured out a good solution for this. The `lib` directory is normally detected relative to the launch directory which for the language server doesn't exist :man_shrugging:

docs/advanced/grouping.md
bentsherman: Never mind, I see this was already updated

PR #601, opened 2025-04-11
Add working with files side quest

adamrtalbot:
Adds a side quest for working with files, vaguely based on the Metadata propagation section of the advanced training.

Story:
 - Create `file` object from string
 - Look at `file` object attributes
 - Extract sample metadata from filename
 - Use `Channel.fromPath` to create a channel of files
 - Extract sample metadata from filename within map operator
 - Use `Channel.fromFilePairs` to create a channel of file pairs
 - Use `publishDir` to save results

Should help introduce the concept of handling files better.

Problems:
 - Doesn't ram home the "always use files as inputs!!!" message. We could do that in the final section on processes?

PR #644, opened 2025-08-05
Intermediate training draft

adamrtalbot:
The content is based on existing training material, that are reordered in this PR + addressing reviews from the original PRs by @adamrtalbot (https://github.com/nextflow-io/training/pull/601) and me (https://github.com/nextflow-io/training/pull/611)

Metadata training:

- Removed filtering to focus on meta maps. filtering operator is covered in working with files / splitting & grouping
- Removed joining channels, instead files are passed through processes
- Fixed typos and language suggestions from original review
- Removed ternary operator, replaced with more verbose if/else statement.
- Moved language on what splitCSV does and what maps are from splitting & grouping
- TODO: How do we want to name this one?

Working with Files

- Shorten the journey to using a meta map
- Remove readNum when using FilePairs, since this is not defined anymore

Splitting & Grouping

- Remove tumor/normal single map early in the training
- Removing section 5.2 & 5.3. We are already using meta maps from the beginning, so this part of the training does not fit into the story anymore.
- TODO: reconsider where we can fit content on collect() operator and groovy collect in.
- TODO: is the ending to abrupt now?

### DIFF
--- a/docs/side_quests/splitting_and_grouping.md
+++ b/docs/side_quests/splitting_and_grouping.md
@@ -97,7 +97,6 @@ We can use the [`splitCsv` operator](https://www.nextflow.io/docs/latest/operato
 
     We'll clarify which one we mean in context, but this distinction is important to understand when working with Nextflow.
 
-
 Apply these changes to `main.nf`:
 
 === "After"


docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
+# Working with sample-specific data
+
+In any scientific analysis, we rarely work with just the raw data files. Each sample comes with its own additional information: what it is, where it came from, and what makes it special. This extra information is what we call sample-specific data.
```
pinin4fjords: Maybe controversial, but I think we should ditch the 'sample' term here, it's overloaded. 

In my head 'sample' has biological meaning relating to laboratories etc. To people from stats it means a random subset of overall dataset.

The concept of 'sample' as something more abstract, a row in a CSV is quite nf-core specific and I don't think it generalises well.

All we need to discuss really is files and metadata.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+│   ├── hallo.txt
+│   ├── hello.txt
+│   ├── hola.txt
+│   ├── salut.txt
+│   └── samplesheet.csv
+├── main.nf
+└── nextflow.config
+```
+
+The samplesheet contains information about different samples and some associated data that we will use in this exercise to tailor our analysis to each sample. Each data files contains greetings in different languages, but we don't know what language they are in. In particular, the samplesheet has 3 columns:
```
pinin4fjords:
```suggestion
The samplesheet contains information about different samples and some associated data that we will use in this exercise to tailor our analysis to each sample. Each data file contains greetings in different languages, but we don't know what language they are in. The samplesheet has 3 columns:
```

docs/side_quests/metadata.md
pinin4fjords: I also think we should sub 'samplesheet' for 'datasheet' in the rest of the material.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+- Understand basic Nextflow concepts (processes, channels, operators)
+
+### 0.2. Starting Point
+
+Let's move into the project directory.
+
+```bash
+cd side-quests/metadata
+```
+
```
pinin4fjords, line 52:
```suggestion

You can set VSCode to focus on this directory:

```bash
code .
```

```

File: docs/side_quests/metadata.md [hunk 4]
```diff
@@ -0,0 +1,732 @@
...
+We'll explore how to handle metadata in workflows. Starting with a simple samplesheet containing basic sample information, you'll learn how to:
+
+- Read and parse sample metadata from CSV files
+- Create and manipulate metadata maps
+- Add new metadata fields during workflow execution
+- Use metadata to customize process behavior
+
+These skills will help you build more robust and flexible pipelines that can handle complex sample relationships and processing requirements.
+
```
pinin4fjords, line 33: We should describe here the problem we'll be solving as an exemplar.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+!!! note
+
+    Throughout this tutorial, we'll use the `ch_` prefix for all channel variables to clearly indicate they are Nextflow channels.
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="3" hl_lines="2-3"
+    ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
+                            .splitCsv(header: true)
+                            .view()
```
pinin4fjords:
```suggestion
    ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
        .splitCsv(header: true)
        .view()
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+    ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
+    ```
+
+We can use the [`splitCsv` operator](https://www.nextflow.io/docs/latest/operator.html#splitcsv) to split the samplesheet into a channel of maps, where each map represents a row from the CSV file.
+
+A map is a key-value data structure similar to dictionaries in Python, objects in JavaScript, or hashes in Ruby. For example:
+
+```groovy
+// Groovy map
+def my_map = [id:'sampleA', character:'squirrel']
```
pinin4fjords, line 132: We can actually have a nextflow file with this content for people to run.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+println my_map.id  // Prints: sampleA
+```
+
+```python
+# Python equivalent dictionary
+my_map = {'id': 'sampleA', 'character': 'squirrel'}
+print(my_map['id'])  # Prints: sampleA
+```
+
```
pinin4fjords:
```suggestion

```

pinin4fjords: Don't think we need this.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+The `header: true` option tells Nextflow to use the first row of the CSV file as the header row, which will be used as keys for the values. Let's see what Nextflow can see after reading with `splitCsv`. To do this, we can use the `view` operator.
```
pinin4fjords:
```suggestion
The `header: true` option tells Nextflow to use the first row of the CSV file as the header row, which will be used as keys for the values. Let's see what Nextflow can see after reading with `splitCsv`. Run the pipeline with the `view()` operator we added above:
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+We can see that each row from the CSV file has been converted into a map with keys matching the header row.
+Each map entry corresponds to a column in our samplesheet:
+
+- `id`
+- `character`
+- `recording`
+
+This format makes it easy to access specific fields from each sample. For example, we could access the sample ID with `id` or the txt file path with `recording`. The output above shows each row from the CSV file converted into a map with keys matching the header row.
+
+Let's access a specific column, the `character` column, that we read in from the samplesheet, and print it. We can use the Nextflow `map` operator to iterate over each `row` in our samplesheet and return specific entry of our map object:
```
pinin4fjords:
```suggestion
Let's access a specific column, the `character` column, that we read in from the samplesheet, and print it. We can use the Nextflow `map` operator to iterate over each item in our channel and return specific entry of our map object:
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="4" hl_lines="2-4"
+                            .splitCsv(header: true)
+                            .map{ row ->
+                              row.character
+                            }
+                            .view()
```
pinin4fjords:
```suggestion
            .splitCsv(header: true)
            .map{ row ->
                row.character
            }
            .view()
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+    ```
+
+=== "Before"
+
+    ```groovy title="main.nf" linenums="4" hl_lines="2-3"
+                            .splitCsv(header: true)
+                            .view()
```
pinin4fjords:
```suggestion
            .splitCsv(header: true)
            .view()
```

docs/side_quests/metadata.md
pinin4fjords: Let's lose the long indents and format as the extension would

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+squirrel
+tux
+sheep
+turkey
+stegosaurus
+moose
+turtle
+```
+
+Success, we can access individual entries or entire rows from our samplesheet.
```
pinin4fjords:
```suggestion
Success, we can use the map structures derived from our sample sheet to access the values from individual columns for each row.
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+Now that we've successfully read in the samplesheet and have access to the data in each row, we can begin implementing our pipeline logic.
+
+### 1.2. Separate meta data and data
+
+In the samplesheet, we have both the input files and data about the input files (`id`, `character`), the meta data.
```
pinin4fjords:
```suggestion
In the samplesheet, we have both the input files and data about the input files (`id`, `character`), the metadata.
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+As we progress through the workflow, we generate more meta data about each sample.
```
pinin4fjords:
```suggestion
As we progress through the workflow, we generate more metadata about each sample.
```

I think we want 'metadata' rather than 'meta data', but we should be consistent in any case. I won't keep marking it, we can do a find/ replace once we agree.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+If we keep all fields as separate Channel elements, it quickly becomes messy: every time we add a new field, we’d have to update every downstream process to expect a different number of inputs, making the code brittle and hard to maintain.
+
+By grouping the metadata into its own key-value map, and keeping the file path as a distinct element, we make our workflow much more robust and flexible. We can add or remove metadata at any stage without having to rewrite process inputs. This approach also keeps process code cleaner and makes it easier to share and reuse code, both within a workflow and across different workflows.
```
pinin4fjords:
````suggestion
So far, every column from the sample sheet has become a separate item in the channel items we derived using `splitCsv()`. Every process that consumes this channel would need to be configured with this structure in mind:

 ```groovy
     input:
     tuple val(id), val(character), file(recording)
``` 

This means that as soon as somebody added an extra column to the sample sheet, the workflow would start producing errors, because the shape of the channel would no longer match what the process expected. It also makes the process hard to share with others who might have slightly different input data, and we end up hard-coding variables into the process that aren't needed by the script block.

To avoid this, we need to find a way of keeping the channel structure consistent however many columns that sample sheet contains, and we can do that by keeping all the metadata in a single part of the channel tuples we call simply the `meta map`.
````

The current text is a bit abstract, so I tried to fix, see what you think.

File: docs/side_quests/metadata.md [hunk 16]
```diff
@@ -0,0 +1,732 @@
...
+
+Let's use this and separate our metadata from the file path. We'll use the `map` operator to restructure our channel elements into a tuple consisting of the meta map and file:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="5" hl_lines="2"
+    .map { row ->
+      [ [id:row.id, character:row.character], row.recording ]
+    }
+    .view()
```
pinin4fjords:
```suggestion
            .map { row ->
                [ [id:row.id, character:row.character], row.recording ]
            }
            .view()
```
Now the indents have gone completely

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+    ```
+
+=== "Before"
+
+    ```groovy title="main.nf" linenums="5"
+    .map{ row ->
+      row.character
+    }
```
pinin4fjords:
```suggestion
            .map{ row ->
                  row.character
            }
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+[[id:sampleA, character:squirrel], /workspaces/training/side-quests/metadata/data/bonjour.txt]
+[[id:sampleB, character:tux], /workspaces/training/side-quests/metadata/data/guten_tag.txt]
+[[id:sampleC, character:sheep], /workspaces/training/side-quests/metadata/data/hallo.txt]
+[[id:sampleD, character:turkey], /workspaces/training/side-quests/metadata/data/hello.txt]
+[[id:sampleE, character:stegosaurus], /workspaces/training/side-quests/metadata/data/hola.txt]
+[[id:sampleF, character:moose], /workspaces/training/side-quests/metadata/data/salut.txt]
+[[id:sampleG, character:turtle], /workspaces/training/side-quests/metadata/data/ciao.txt]
+```
+
+We have successfully separated our values into their own map to separate it from the file. Each of our channel elements now has the shape `[ meta,  file ]`. Each process that now uses the data can be of similar input shape even if we keep adding or removing fields.
```
pinin4fjords:
````suggestion
Now, each item in the channel has only two items, the metadata map (e.g. `[id:sampleA, character:squirrel]`) and the data file described by that metadata (e.g. `/workspaces/training/side-quests/metadata/data/bonjour.txt`).

Now we can write processes to consume the channel without hard-coding the metadata items into the input specification:

 ```groovy
     input:
     tuple val(meta), file(recording)
``` 

Additional columns in the sample sheet will make more properties available in the `meta` map, but won't change the channel shape.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+- **How to read in a samplesheet**: Using `splitCsv` to read CSV files with header information and transform rows into structured data
+- **How to create a meta map**: Separating metadata from file data using tuple structure `[ [id:value, ...], file ]`
+
+---
+
+## 2. Create new meta map keys
+
+### 2.1. Passing the meta map through a process
+
+Now we want to process our samples with unidentified languages. Let's add a process definition before the `workflow` that can identify the language in each file:
```
pinin4fjords: This is the first time we've heard that there are unidentified languages (see my comment in the intro re: a problem statement)

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+You may have noticed something else: we kept the meta data of our samples and associated it with our new piece of information. We achieved this by adding the `meta` map to the output tuple in the process:
+
+```groovy title="main.nf" linenums="12"
+output:
+      tuple val(meta), path(file), stdout
+```
+
+This is a useful way to ensure the sample-specific information stays connected with any new information that is generated.
+
```
pinin4fjords, line 381: Could we extend the material slightly to accommodate using metadata to join channels, and pass to a process that consumes the joint channel? Would be nice to incorporate this, discuss the anti-pattern of monster channels, and highlight it as a main mechanism of getting round the lack of order in Nextflow channels.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+After running the language prediction, each element in the output channel looks like this:
+
+```console
+[meta, file, lang]  // e.g. [[id:sampleA, character:squirrel], bonjour.txt, fr]
+```
+
+The `map` operator takes each channel element and processes it to create a modified version. Inside the closure `{ meta, file, lang -> ... }`, we then take the existing `meta` map, create a new map `[lang:lang]`, and merge both together using `+`.
+
+The `+` operator in Groovy merges two maps together. So if our original `meta` was `[id:sampleA, character:squirrel]`, then `meta + [lang:'fr']` creates a new map: `[id:sampleA, character:squirrel, lang:fr]`.
```
pinin4fjords:
````suggestion

The `map` operator takes each channel element and processes it to create a modified version. Inside the closure `{ meta, file, lang -> ... }`, we then take the existing `meta` map, create a new map `[lang:lang]`, and merge both together using `+`.

The `+` operator in Groovy merges two maps together. So if our original `meta` was `[id:sampleA, character:squirrel]`, then `meta + [lang:'fr']` creates a new map: `[id:sampleA, character:squirrel, lang:fr]`.

!!! Note

    The `+` notation with maps creates an entirely new map object, which is what we want. If we'd done something like `meta.lang = lang` we'd have been modifying the original object, which can lead to unpredictable effects. 

```bash title="View new meta map key"
nextflow run main.nf -resume
```

```console title="View new meta map key"

 N E X T F L O W   ~  version 24.10.4

Launching `main.nf` [cheeky_fermat] DSL2 - revision: d096281ee4

[4e/f722fe] IDENTIFY_LANGUAGE (7) [100%] 7 of 7, cached: 7 ✔
[[id:sampleA, character:squirrel, lang:fr], /workspaces/training/side-quests/metadata/work/eb/f7148ebdd898fbe1136bec6a714acb/bonjour.txt]
[[id:sampleB, character:tux, lang:de], /workspaces/training/side-quests/metadata/work/16/71d72410952c22cd0086d9bca03680/guten_tag.txt]
[[id:sampleC, character:sheep, lang:de], /workspaces/training/side-quests/metadata/work/ea/04f5d979429e4455e14b9242fb3b45/hallo.txt]
[[id:sampleD, character:turkey, lang:en], /workspaces/training/side-quests/metadata/work/c4/b7562adddc1cc0b7d414ec45d436eb/hello.txt]
[[id:sampleF, character:moose, lang:fr], /workspaces/training/side-quests/metadata/work/5a/6c2b84bf8fadb98e28e216426be079/salut.txt]
[[id:sampleE, character:stegosaurus, lang:es], /workspaces/training/side-quests/metadata/work/af/ee7c69bcab891c40d0529305f6b9e7/hola.txt]
[[id:sampleG, character:turtle, lang:it], /workspaces/training/side-quests/metadata/work/4e/f722fe47271ba7ebcd69afa42964ca/ciao.txt]
```

Nice, we expanded our meta map with new information we gathered in the pipeline. After running the language prediction, each element in the output channel looks like this:

```console
[meta, file, lang]  // e.g. [[id:sampleA, character:squirrel], bonjour.txt, fr]
```
````

I found myself wanting to understand what was going on before I ran the command.

I'd also like to use this as an excuse to state Rob's lessons about not modifying maps in-place.

File: docs/side_quests/metadata.md [hunk 22]
```diff
@@ -0,0 +1,732 @@
...
+
+<!-- TODO Should we also show how to remove a key using subMap?! -->
+
+### 2.3. Assign a language group using conditionals
+
+Alright, now that we have our language predictions, let's use the information to assign them into new groups. In our example data, we have provided data sets that belong either to `germanic` (either English or German) or `romance` (French, Spanish, Italian) languages.
```
pinin4fjords:
```suggestion
Now that we have our language predictions, let's use the information to assign them into new groups. In our example data, we have provided data sets that belong either to `germanic` (either English or German) or `romance` (French, Spanish, Italian) languages.
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+We can use the `map` operator to assign either group.
```
pinin4fjords:
```suggestion
We can add another `map` operator to assign either group (add this below your last map operation).
```

I was confused about whether I should edit the map from the previous example

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+In this section, you've learned:
+
+- **Why metadata is important**: Keeping metadata with your data preserves important sample information throughout the workflow.
+- **How to read in a samplesheet**: Using `splitCsv` to read CSV files with header information and transform rows into structured data
+- **How to create a meta map**: Separating metadata from file data using tuple structure `[ [id:value, ...], file ]`
+
+---
+
+## 2. Create new meta map keys
```
pinin4fjords:
```suggestion
## 2. Manipulating metadata
```

I think the content of this section suggests a more generic title

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+### 2.1. Passing the meta map through a process
```
pinin4fjords:
```suggestion
### 2.1. Copying input metadata to output channels
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+You may have noticed something else: we kept the meta data of our samples and associated it with our new piece of information. We achieved this by adding the `meta` map to the output tuple in the process:
+
+```groovy title="main.nf" linenums="12"
+output:
+      tuple val(meta), path(file), stdout
+```
+
+This is a useful way to ensure the sample-specific information stays connected with any new information that is generated.
+
+### 2.2. Add the language prediction to the meta map
```
pinin4fjords:
```suggestion
### 2.2. Using process outputs to augment metadata
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+```
+
+Let's understand how this transformation works. The `map` operator here again takes a closure that processes each element in the channel. Inside the closure, we're using an if-clause to create a new language group classification.
+
+Here's what's happening step by step:
+
+- Create a new field `lang_group`: We set a default to `romance`
+- Extract existing metadata: We access `meta.lang` (the language we predicted earlier) from the existing meta map
+- Apply conditional logic: We use an if-clause to determine the language group based on the language: Is `meta.lang` either `de` or `en`, then we re-assign `lang_group` to `germanic`
+- Merge with existing metadata: We use `meta + [lang_group:lang_group]` to combine the existing meta map with our new field
```
pinin4fjords:
```suggestion
- Merge with existing metadata: We use `meta + [lang_group:lang_group]` in the same way as before to combine the existing meta map with our new field
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+The resulting channel elements maintain their `[meta, file]` structure, but the meta map now includes this new classification.
+
+### Takeaway
+
+In this section, you've learned how to :
+
+- **Create custom keys**: You created two new keys in your meta map, merging them with `meta + [new_key:value]` into the existing meta map. One based on a computed value from a process, and one based on a condition you set in the `map` operator.
```
pinin4fjords, line 511:
```suggestion
- **Apply input metadata to output channels**: Copying metadata in this way allows us to associate results later on based on metadata content.
- **Create custom keys**: You created two new keys in your meta map, merging them with `meta + [new_key:value]` into the existing meta map. One based on a computed value from a process, and one based on a condition you set in the `map` operator.
```

Missing a point

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+These allow you to associated new and existing meta data with files as you progress through your pipeline.
+
+---
+
+## 3. Customize a process with meta map
+
+Let's let some fun characters say the phrases that we have passed in. In the [hello-nextflow training](../hello_nextflow/05_hello_containers.md), you already encountered the `cowpy` package, a python implementation of a tool called `cowsay` that generates ASCII art to display arbitrary text inputs in a fun way. We will re-use a process from there.
```
pinin4fjords:
```suggestion
Let's make some fun characters say the phrases from the files our channel. In the [hello-nextflow training](../hello_nextflow/05_hello_containers.md), you already encountered the `cowpy` package, a python implementation of a tool called `cowsay` that generates ASCII art to display arbitrary text inputs in a fun way. We will re-use a process from there.
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+In this section, you've learned how to :
+
+- **Create custom keys**: You created two new keys in your meta map, merging them with `meta + [new_key:value]` into the existing meta map. One based on a computed value from a process, and one based on a condition you set in the `map` operator.
+
+These allow you to associated new and existing meta data with files as you progress through your pipeline.
+
+---
+
+## 3. Customize a process with meta map
```
pinin4fjords:
```suggestion
## 3. Use meta map properties in a process
```

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+Copy in the process before your workflow block:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="20"
+    /*
+     * Generate ASCII art with cowpy
+    */
+    process COWPY {
+
```
pinin4fjords, line 530:
```suggestion
        publishDir "results/${meta.lang_group}", mode: 'copy'
```

Let's just include the publishDir in the process to begin with, since it's not a teaching point here.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+=== "Before"
+
+    ```groovy title="main.nf" linenums="20"
+    workflow{
+      ...
+    }
+    ```
+
+### 3.1. Add a custom publishing location
```
pinin4fjords:
```suggestion
### 3.1. Use meta map information in the process definition
```

That's the key point here- rather than the publishDir itself

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
+=== "Before"
+
+    ```groovy title="main.nf" linenums="59"
+                                      [ meta + [lang_group:lang_group], file ]
+                                  }
+                                  .view()
+    ```
+
+We are still missing a publishing location. Given we have been trying to figure out what languages our samples were in, let's group the samples by language in the output directory. Earlier, we added the predicted language to the `meta` map. We can access this `key` in the process and use it in the `publishDir` directive:
```
pinin4fjords:
```suggestion
We also need to add a publishing location, so that we can view our comedy creatures saying our phrases. Given we have been trying to figure out what languages our samples were in, let's group the samples by language in the output directory. 

Because we added the predicted language to the `meta` map, it's available for us to use in the process definition, and we can use it here in our process directive:
```

Just trying to tweak emphasis so that it's clear that the teaching point is that you can use the metadata in the process (rather than something about publishing itself). Not sure I've quite hit the mark yet...

docs/side_quests/metadata.md
pinin4fjords: Are they properties or attributes? I don't remember we we should use the right one

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+  \
+   \   \_\_    _/_/
+    \      \__/
+           (oo)\_______
+           (__)\       )\/\
+               ||----w |
+               ||     ||
+```
+
+This is a subtle difference to other parameters that we have set in the pipelines in previous trainings. A parameter that is passed as part of the `params` object is generally applied to all samples. When a more surgical approache is necessary, using the sample specific information is a good alternative.
```
pinin4fjords:
```suggestion

```

I'm not sure I understand the association with `params`.

docs/side_quests/metadata.md
```diff
@@ -0,0 +1,732 @@
...
+
```
pinin4fjords, line 712:
````suggestion

This example shows you what's possible when you have a meta map, but it's worth noting a limitation of this approach. By using a property of the meta map in the script block, we introduce a hard requirement on the properties that must be present. Anyone running with a sample sheet that did not contain the `character` property would encounter an error. We could set workflow-level validation to deal with that, but it's also better practice to define the process with such hard requirements as explicit inputs:

```groovy
    input:
    tuple val(meta), val(character), path(input_file)
```

... then we extract the metadata property at the workflow level before running the process:

```groovy
        COWPY(ch_languages.map{meta, file -> [meta, meta.character, file] )
```

This way, if we fail to provide the expected input to the process, Nextflow will be much clearer with us what the problem is.
````

I know it's a bit of an nf-core thing, but there are really good reasons for it and I don't want to encourage bad patterns.

**Review by pinin4fjords — COMMENTED** [review]
Ok, here's my review of the metadata one. Main points:

- I'd like to excise 'sample' as a thing here
- Need to tidy up indentation, they're a bit mad right now
- Could we benefit from a section on combining channels using metas?
- Let's try not to encourage bad process designe?

**Comment by FriederikeHanssen** [comment]
> Could we benefit from a section on combining channels using metas?

we had this and then I removed it 🫠 https://github.com/FriederikeHanssen/training/blob/54c9d4c347d3c1db2171a0b55dd3bccbe3e7ccb0/docs/side_quests/metadata.md?plain=1#L323  reason being, that combining and grouping is covered in the splitting_grouping training and I was worried this one would be overloaded with concepts.

**Comment by pinin4fjords** [comment]
> > Could we benefit from a section on combining channels using metas?
> 
> we had this and then I removed it 🫠 https://github.com/FriederikeHanssen/training/blob/54c9d4c347d3c1db2171a0b55dd3bccbe3e7ccb0/docs/side_quests/metadata.md?plain=1#L323 reason being, that combining and grouping is covered in the splitting_grouping training and I was worried this one would be overloaded with concepts.

OK, fair concern, but I think we should at least cover it in outline, and maybe link to that other material. It's not obvious when you start out that joining on maps is even possible, and for me it's one of the main reasons to add meta maps to channels.

File: docs/side_quests/working_with_files.md [hunk 36]
```diff
@@ -0,0 +1,1069 @@
...
+- Basic understanding of using sample-specific data (meta data): [Working with sample-specific data](./metadata.md)
+
+### 0.2. Starting Point
+
+Let's move into the project directory:
+
+```bash
+cd side-quests/working_with_files
+```
+
```
pinin4fjords, line 37: Do the `code .` thing

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+│   ├── sampleB_rep1_tumor_R1_001.fastq.gz
+│   ├── sampleB_rep1_tumor_R2_001.fastq.gz
+│   ├── sampleC_rep1_normal_R1_001.fastq.gz
+│   ├── sampleC_rep1_normal_R2_001.fastq.gz
+│   ├── sampleC_rep1_tumor_R1_001.fastq.gz
+│   └── sampleC_rep1_tumor_R2_001.fastq.gz
+└── main.nf
+```
+
+This directory contains paired-end sequencing data for three samples (A, B, C), with the typical `_R1_` and `_R2_` naming convention for forward and reverse reads. Each sample has normal and tumor tissue types, and sample A has two replicates.
```
pinin4fjords:
```suggestion
This directory contains DNA sequencing data for three samples (A, B, C). Every sample has data from both normal and tumour tissue, and sample A has an additional replicate for both of those (maybe the authors had to sequence twice to get enough data). This is a type of sequencing data called 'paired end', which means that each data point consists of two files  with an `_R1_` and `_R2_` naming convention.
```

Similar to the metadata one, I'm confused by the 'sample' usage here. I would think that e.g. `sampleC_rep1_normal` would constitute a 'sample' for example, and would be the thing that mapped to a single row in a sample sheet. A, B, C would be 'patient', maybe?

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+```console title="File object output"
+ N E X T F L O W   ~  version 24.10.4
+
+Launching `main.nf` [kickass_coulomb] DSL2 - revision: 5af44b1b59
+
+/workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz
+```
+
+You should only see a single difference, the file path will now be absolute. Note that the full path will change based on where you are doing this training, but it is likely to be something like `/workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz`.
+
```
pinin4fjords, line 158: I would actually move the 'remote files' bit you have at the end to up here. I don't think it should be an afterthought, it should up-front as native thing, like 'Nextflow works exactly the same with local files and remote ones'.

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+### 1.2. File Attributes
+
+Why is this helpful? Well now Nextflow understands that `myFile` is a file object and not just a string, we can access the various attributes of the file object.
+
+Let's update our workflow to print out the file attributes:
```
pinin4fjords, line 163:
```suggestion
Let's update our workflow to print out some path attributes:
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+Launching `main.nf` [kickass_coulomb] DSL2 - revision: 5af44b1b59
+
+/workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz
+```
+
+You should only see a single difference, the file path will now be absolute. Note that the full path will change based on where you are doing this training, but it is likely to be something like `/workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz`.
+
+### 1.2. File Attributes
+
+Why is this helpful? Well now Nextflow understands that `myFile` is a file object and not just a string, we can access the various attributes of the file object.
```
pinin4fjords:
```suggestion
Why is this helpful? Well now Nextflow understands that `myFile` is a [Path object](https://www.nextflow.io/docs/latest/reference/stdlib-types.html#path) and not just a string, we can access the various attributes of the path object.
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+data/sampleA_rep1_normal_R1_001.fastq.gz
+```
+
+---
+
+## 1. Basic File Operations
+
+### 1.1. Creating File Objects
+
+Let's start by understanding how to create file objects in Nextflow. In our workflow, we have a string path `data/sampleA_rep1_normal_R1_001.fastq.gz`. This is just a plain string - Nextflow doesn't automatically recognize it as representing a file. To work with files properly in Nextflow, we need to convert string paths into proper file objects using the `file()` method, which provides access to file properties and operations.
```
pinin4fjords:
````suggestion
So far, so good, but currently the path we're looking at is just a string as far as Nextflow is concerned, no different to 'hello' or 'supercalafragalisticexpialodocious'. We can actually confirm that by checking the type:

=== "After"

    ```groovy title="main.nf" linenums="2" hl_lines="2"
    // Create a file object from a string path
    myFile = file('data/sampleA_rep1_normal_R1_001.fastq.gz')
    println "${myFile} is of type ${myFile.class}"
    ```

=== "Before"

    ```groovy title="main.nf" linenums="2" hl_lines="2"
    // Create a file object from a string path
    myFile = 'data/sampleA_rep1_normal_R1_001.fastq.gz'
    println "${myFile}"
    ```

Run the workflow:

```bash
nextflow run main.nf
```

```console title="File object output"
 N E X T F L O W   ~  version 25.04.3

Launching `main.nf` [gloomy_euler] DSL2 - revision: 6c02ad9221

data/sampleA_rep1_normal_R1_001.fastq.gz is of type class java.lang.String
```


If we were to pass that to a process and attempt to read the contents of the file we'd encounter problems.

// Maybe actually do that and show problems?

Nextflow has a special concept for dealing with all paths such as local files and directories, HTTP/FTP URLs, and object storage paths (e.g. Amazon S3), called, unsurprisingly, '[Path](https://www.nextflow.io/docs/latest/reference/stdlib-types.html#path)'. If we pass a path object to a Nextflow process, Nextflow will stage the file to the process working directory, so that we can work with it. 

So, let's start by understanding how to create Path objects in Nextflow. In our workflow, we have a string path `data/sampleA_rep1_normal_R1_001.fastq.gz`, which we need to convert to a Path, which we can do with the `file()` method.
````

Comment (pinin4fjords) [reply docs/side_quests/working_with_files.md]: We might as well be precise - `file()` makes Path objects (at least that's what the docs say).

File: docs/side_quests/working_with_files.md [hunk 42]
```diff
@@ -0,0 +1,1069 @@
...
+
```
pinin4fjords, line 115: I wonder if maybe we should have an example here where we attempt to do something file-ey with a string object in a process.

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+File name: sampleA_rep1_normal_R1_001.fastq.gz
+Simple name: sampleA_rep1_normal_R1_001
+Extension: gz
+Parent directory: /workspaces/training/side-quests/working_with_files/data
+```
+
+When using a path as a string, Nextflow has no idea what it's looking at - it's just a series of characters. When we use the `file()` method, Nextflow understands this is a file and can access its properties such as name, extension, and parent directory. This also tells Nextflow exactly how to handle it in the workflow, which prevents common errors.
+
+Despite their prevalence in bioinformatics, files are not strings! Nextflow needs to know it's working with a file object to properly manage it throughout your workflow.
```
pinin4fjords:
```suggestion
Because we built a `Path` object we now have convenient access to a bunch of attributes- see [the documentation](https://www.nextflow.io/docs/latest/reference/stdlib-types.html#path) for the full set.
```

If we cover this up-front per my suggestion above

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+### Takeaway
+
+- The `file()` method creates a Path object from a string, which Nextflow understands as a file
+- You can access file properties like `name`, `simpleName`, `extension`, and `parent` [using file attributes](https://www.nextflow.io/docs/latest/working-with-files.html#getting-file-attributes)
+- Using file objects instead of strings allows Nextflow to properly manage files in your workflow
```
pinin4fjords:
```suggestion
- Using Path objects instead of strings allows Nextflow to properly manage files in your workflow
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+---
+
+## 2. Using Channels for File Handling
```
pinin4fjords:
```suggestion
## 2. Reading files using the `fromPath()` channel factory
```
This seems to be what the section is really about

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+While the `file()` method is useful for simple file operations, channels provide a more powerful way to handle multiple files in workflows. We could load our file into a channel using [`Channel.of()`](https://www.nextflow.io/docs/latest/reference/channel.html#of), but we have a much more convenient tool called [`Channel.fromPath()`](https://www.nextflow.io/docs/latest/reference/channel.html#frompath) which finds every file in a directory that matches a given pattern.
```
pinin4fjords:
````suggestion
The `file()` method is useful for simple file operations, and we can combine that with [`Channel.of()`](https://www.nextflow.io/docs/latest/reference/channel.html#of) to build channels from files like:

```groovy title="Channel.of() with file()"
    ch_fastq = Channel.of([file('data/sampleA_rep1_normal_R1_001.fastq.gz')])
```

But we have a much more convenient tool called [`Channel.fromPath()`](https://www.nextflow.io/docs/latest/reference/channel.html#frompath) which generates a channel from static file strings as well as glob patterns.
````

File: docs/side_quests/working_with_files.md [hunk 47]
```diff
@@ -0,0 +1,1069 @@
...
+
+### 2.1. Reading Files with Channel.fromPath
+
+Update your `main.nf` file:
```
pinin4fjords:
```suggestion
Update your `main.nf` file to use `fromPath` with your single file:
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+    myFile = file('data/sampleA_rep1_normal_R1_001.fastq.gz')
+    println "${myFile}"
+    ```
+
+=== "Before"
+
+    ```groovy title="main.nf" linenums="2" hl_lines="2"
+    // Create a file object from a string path
+    myFile = 'data/sampleA_rep1_normal_R1_001.fastq.gz'
+    println "${myFile}"
```
pinin4fjords:
```suggestion
    println "${myFile} is of type ${myFile.class}"
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+Let's start by understanding how to create file objects in Nextflow. In our workflow, we have a string path `data/sampleA_rep1_normal_R1_001.fastq.gz`. This is just a plain string - Nextflow doesn't automatically recognize it as representing a file. To work with files properly in Nextflow, we need to convert string paths into proper file objects using the `file()` method, which provides access to file properties and operations.
+
+Edit the `main.nf` file to include the following:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="2" hl_lines="2"
+    // Create a file object from a string path
+    myFile = file('data/sampleA_rep1_normal_R1_001.fastq.gz')
+    println "${myFile}"
```
pinin4fjords:
```suggestion
    println "${myFile} is of type ${myFile.class}"
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+```console title="File object output"
+ N E X T F L O W   ~  version 24.10.4
+
+Launching `main.nf` [kickass_coulomb] DSL2 - revision: 5af44b1b59
+
+/workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz
+```
+
+You should only see a single difference, the file path will now be absolute. Note that the full path will change based on where you are doing this training, but it is likely to be something like `/workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz`.
```
pinin4fjords:
```suggestion
You should see that the class is now `sun.nio.fs.UnixPath` and the file path will now be absolute. Note that the full path will change based on where you are doing this training, but it is likely to be something like `/workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz`.
```

docs/side_quests/working_with_files.md
pinin4fjords: we could also show that you get a `nextflow.file.http.XPath` back from `file()` with URIs, but you can treat it the same as the UnixPath.

docs/side_quests/working_with_files.md
pinin4fjords: I'm not going to carry on correcting this- I can do that if we agree it's a good plan

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+```console title="Channel.fromPath Output"
+ N E X T F L O W   ~  version 24.10.4
+
+Launching `main.nf` [grave_meucci] DSL2 - revision: b09964a583
+
+Found file: /workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz
+```
+
+Note how Nextflow has grabbed the file we specified and turned it into a `file` object. `Channel.fromPath()` is a convenient way of creating a new channel populated by a list of files.
```
pinin4fjords:
```suggestion
Note how Nextflow has grabbed the file we specified and turned it into a `Path` type object, in exactly the same way that `file()` would have done. `Channel.fromPath()` is just a convenient way of creating a new channel populated by a list of files.
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+    ```
+
+=== "Before"
+
+    ```groovy title="main.nf" linenums="2" hl_lines="5-8"
+    // Create a file object from a string path
+    myFile = file('data/sampleA_rep1_normal_R1_001.fastq.gz')
+
+    // Print file attributes
+    println "File name: ${myFile.name}"
```
pinin4fjords:
```suggestion
    println "File name: ${myFile.name}"
    println "File class: ${myFile.class}"
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+Let's update our workflow to print out the file attributes:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="2" hl_lines="5-9"
+    // Create a file object from a string path
+    myFile = file('data/sampleA_rep1_normal_R1_001.fastq.gz')
+
+    // Print file attributes
+    println "File name: ${myFile.name}"
```
pinin4fjords:
```suggestion
    println "File name: ${myFile.name}"
    println "File class: ${myFile.class}"
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+```
+
+You'll see various file attributes printed to the console:
+
+```console title="File Attributes Output"
+ N E X T F L O W   ~  version 24.10.4
+
+Launching `main.nf` [ecstatic_ampere] DSL2 - revision: f3fa3dcb48
+
+File name: sampleA_rep1_normal_R1_001.fastq.gz
```
pinin4fjords:
```suggestion
File name: sampleA_rep1_normal_R1_001.fastq.gz
File class: sun.nio.fs.UnixPath
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+- `Channel.fromPath()` creates a channel with files matching a pattern
+- Each file is emitted as a separate element in the channel
+- We can use a glob pattern to match multiple files
+
+---
+
+## 3. Extracting Sample Metadata from Filenames
+
+One of the most common tasks in bioinformatics workflows is extracting metadata from filenames. This is especially important when working with sequencing data, where filenames often contain information about the sample, condition, replicate, and read number.
```
pinin4fjords:
```suggestion
One of the most common tasks in bioinformatics workflows is extracting metadata from filenames. This is usually feasible when working with sequencing data, where filenames often contain information about the sample, condition, replicate, and read number.
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+[sampleA_rep1_normal_R2_001, /workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R2_001.fastq.gz]
+```
+
+Note how we have separated the sample `simpleName`, which includes the metadata, from the `file` object. This is useful if we want to use the sample metadata in a later process.
+
+### 3.2. Extracting Metadata from Filenames
+
+Our metadata is embedded in the filename, but it's not in a standard format. We need to split up the filename into it's components which are separated by underscores.
+
+Nextflow includes a method called `tokenize()` which is perfect for this task.
```
pinin4fjords:
```suggestion
Groovy includes a method called `tokenize()` which is perfect for this task.
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+    ```groovy title="main.nf" linenums="4" hl_lines="3-11"
+    ch_fastq.map { myFile ->
+        def (sample, replicate, type, readNum) = myFile.simpleName.tokenize('_')
+        [
+          [
+            id: sample,
+            replicate: replicate,
+            type: type,
+            readNum: readNum,
```
pinin4fjords:
```suggestion
            readNum: readNum.replace('rep', ''),
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="4" hl_lines="3-11"
+    ch_fastq.map { myFile ->
+        def (sample, replicate, type, readNum) = myFile.simpleName.tokenize('_')
+        [
+          [
+            id: sample,
+            replicate: replicate,
```
pinin4fjords:
```suggestion
            replicate: replicate.replace('rep', ''),
```

pinin4fjords: Let's just include these up front, text munging isn't a main teaching outcome here

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+```console title="Simplified Metadata Output"
+ N E X T F L O W   ~  version 24.10.4
+
+Launching `main.nf` [reverent_volta] DSL2 - revision: b3aac71fea
+
+[[sample:sampleA, replicate:1, type:normal, readNum:2], /workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R2_001.fastq.gz]
+[[sample:sampleA, replicate:1, type:normal, readNum:1], /workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz]
+```
+
+Nice! we have removed the "rep" from the replicate string.
```
pinin4fjords:
```suggestion
```

Don't think this deserves a separate subsection, not telling us anything new about file handling

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+```
+
+Notice, that this time we don't have a `readNum`. Because read 1 and read 2 are kept together, we do not need to track this in the meta data.
+
+Well done! We have grabbed the metadata from the filenames and used them as values in the tuple.
+
+### Takeaway
+
+- [`Channel.fromFilePairs()` automatically finds and pairs related files](https://www.nextflow.io/docs/latest/reference/channel.html#fromfilepairs)
+- The closure defines how to extract sample IDs from filenames
```
pinin4fjords: We didn't cover anything about usage of the closure, did we?

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+│   └── sampleB_stats.txt
+└── sampleC
+    └── sampleC_stats.txt
+```
+
+See how we have analyzed all the samples in one go!
+
+Wait, we have a problem. We have 2 replicates for sampleA, but only 1 output file! We are overwriting the output file each time.
+
+### 5.4. Make the published files unique
```
pinin4fjords: This is a duplicated point from the metdata module, but I think that's OK, it pertains to both metadata and file handling.

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+            └── sampleC_stats.txt
+```
+
+See how using sample metadata as values gives us powerful flexibility in our pipeline. By propagating metadata alongside our data in tuples, we can:
+
+1. Create organized output directories based on sample attributes
+2. Make decisions in processes based on sample properties
+3. Split, join, and recombine data based on metadata values
+
+This pattern of keeping metadata explicit and attached to the data (rather than encoded in filenames) is a core best practice in Nextflow that enables building robust, maintainable bioinformatics workflows.
```
pinin4fjords: I do feel like some of this should be replaced with a link to the metadata material.

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+### Takeaway
+
+- Processes can take files as input and produce new files as output
+- The `publishDir` directive organizes outputs based on metadata values
```
pinin4fjords:
```suggestion
- The `publishDir` directive can organize outputs based on metadata values
```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+1. Create organized output directories based on sample attributes
+2. Make decisions in processes based on sample properties
+3. Split, join, and recombine data based on metadata values
+
+This pattern of keeping metadata explicit and attached to the data (rather than encoded in filenames) is a core best practice in Nextflow that enables building robust, maintainable bioinformatics workflows.
+
+### Takeaway
+
+- Processes can take files as input and produce new files as output
```
pinin4fjords:
```suggestion
```
Don't think that's a main teaching point here

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+- The `publishDir` directive organizes outputs based on metadata values
+- Metadata in tuples enables structured organization of results
+- This approach creates maintainable workflows with clear data provenance
+
+---
+
+## 6. Using remote files
```
pinin4fjords: As a said further up, I don't think this should be a 'by the way' at the end, but rather something we show up-front.

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1069 @@
...
+
+!!! warning
+Accessing remote data requires an internet connection!
+
+In your workflow, replace the string path with an HTTPS one to download this file from the internet. We are going to swap the relative path of the FASTQ files with the remote one stored on the internet. This is the same data as we have been previously using.
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="27"
+    ch_fastq = Channel.fromFilePairs('https://github.com/nextflow-io/training/blob/bb187e3bfdf4eec2c53b3b08d2b60fdd7003b763/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz')
```
pinin4fjords: This would need to be a glob path, no?

pinin4fjords: No, actually, we'd need to not use `fromFilePairs`, right?

pinin4fjords (commented):
This is my review for the file handling module. Bigger points:

- Remote files up-front
- Clarify Path objects, use `.class` to demonstrate
- Don't over-egg things that aren't the main teaching points here (string parsing, metadata handling)

docs/side_quests/working_with_files.md
adamrtalbot: It works fine as it is. Using a glob (`*` or `{}`) didn't work, but just using a string worked fine.

pinin4fjords: Yeah, but it only produces one file:

```
[sampleA_rep1_normal_R1_001, [/nextflow-io/training/blob/bb187e3bfdf4eec2c53b3b08d2b60fdd7003b763/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz]]
```

Somewhat defeats the point of `fromFilePairs`, it's only really useful with the glob.

docs/side_quests/splitting_and_grouping.md
```diff
@@ -120,26 +121,73 @@ Launching `main.nf` [deadly_mercator] DSL2 - revision: bd6b0224e9
...
+
+    For a more extensive introduction on working with sample-specific data, you can work through the training [Sample Specific Data](./metadata.md)
+
+Let's combine the sample meta data into its own map, separating it from the files:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="2" hl_lines="3-5"
+    ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
+                        .splitCsv(header: true)
```
pinin4fjords: Would be good to make the indents sane in this module

docs/side_quests/splitting_and_grouping.md
```diff
@@ -262,39 +317,31 @@ nextflow run main.nf
...
 
-    ```groovy title="main.nf" linenums="2"
-    ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
-                        .splitCsv(header: true)
-    ch_normal_samples = ch_samplesheet
-                        .filter { sample -> sample.type == 'normal' }
+    ```groovy title="main.nf" linenums="10"
     ch_tumor_samples = ch_samplesheet
-                        .filter { sample -> sample.type == 'tumor' }
+                        .filter { meta, file -> meta.type == 'tumor' }
```
pinin4fjords: I think we're belabouring the point a little with the view things here, `view` shouldn't be a major learning outcome here, it was covered in 'Hello'. We should just use closures with the views from the start to label normal and tumour.

docs/side_quests/splitting_and_grouping.md
```diff
@@ -347,40 +394,36 @@ nextflow run main.nf
 
 Launching `main.nf` [loving_bardeen] DSL2 - revision: 012d38e59f
 
-[id:sampleA, repeat:1, type:tumor, bam:sampleA_rep1_tumor.bam]
-[id:sampleA, repeat:2, type:tumor, bam:sampleB_rep1_tumor.bam]
-[id:sampleB, repeat:1, type:tumor, bam:sampleC_rep1_tumor.bam]
-[id:sampleC, repeat:1, type:tumor, bam:sampleD_rep1_tumor.bam]
+[[id:sampleA, repeat:1, type:tumor], sampleA_rep1_tumor.bam]
```
pinin4fjords: (sorry, commenting at the closest change)

we should make clear that the element used for `join` is configurable

docs/side_quests/splitting_and_grouping.md
```diff
@@ -120,26 +121,73 @@ Launching `main.nf` [deadly_mercator] DSL2 - revision: bd6b0224e9
...
+
+We separated the sample meta data from the files into a map.
+We now have a channel of maps and files, each representing a row from the samplesheet, which we will use in this training to split and group our workload.
 
 ### Takeaway
 
 In this section, you've learned:
 
 - **Reading in a samplesheet**: How to read in a samplesheet with `splitCsv`
+- **Combining sample-specific information**: Using groovy maps to hold information about a sample
```
pinin4fjords: Probably remove `.view()` as a learning outcome here

docs/side_quests/splitting_and_grouping.md
```diff
@@ -308,10 +355,10 @@ nextflow run main.nf
...
 Launching `main.nf` [loving_bardeen] DSL2 - revision: 012d38e59f
 
-[id:sampleA, repeat:1, type:tumor, bam:sampleA_rep1_tumor.bam]
-[id:sampleA, repeat:2, type:tumor, bam:sampleB_rep1_tumor.bam]
-[id:sampleB, repeat:1, type:tumor, bam:sampleC_rep1_tumor.bam]
-[id:sampleC, repeat:1, type:tumor, bam:sampleD_rep1_tumor.bam]
+[[id:sampleA, repeat:1, type:tumor], sampleA_rep1_tumor.bam]
+[[id:sampleA, repeat:2, type:tumor], sampleA_rep2_tumor.bam]
+[[id:sampleB, repeat:1, type:tumor], sampleB_rep1_tumor.bam]
+[[id:sampleC, repeat:1, type:tumor], sampleC_rep1_tumor.bam]
```
pinin4fjords: Remove `view()` as a learning outcome below

docs/side_quests/splitting_and_grouping.md
```diff
@@ -925,17 +1018,17 @@ We can reuse the `subMap` method from before to isolate our `id` and `interval`
 
 === "After"
 
-    ```groovy title="main.nf" linenums="25" hl_lines="10-17"
+    ```groovy title="main.nf" linenums="28" hl_lines="11-20"
```
pinin4fjords:
```suggestion
    ```groovy title="main.nf" linenums="28" hl_lines="10-19"
```

docs/side_quests/splitting_and_grouping.md
```diff
@@ -995,27 +1088,29 @@ Let's now group the samples by this new grouping element, using the [`groupTuple
...
+[
+grouping_key.subMap('id', 'interval'),
+normal,
+tumor
+]
+}
+.groupTuple()
+.view()
+
+````
```
pinin4fjords:
````suggestion
```
````

fix syntax error

File: docs/side_quests/splitting_and_grouping.md [hunk 74]
```diff
@@ -925,17 +1018,17 @@ We can reuse the `subMap` method from before to isolate our `id` and `interval`
 
 === "After"
 
-    ```groovy title="main.nf" linenums="25" hl_lines="10-17"
+    ```groovy title="main.nf" linenums="28" hl_lines="11-20"
     ch_combined_samples = ch_joined_samples.combine(ch_intervals)
-                          .map { grouping_key, normal, tumor, interval ->
+                          .map { grouping_key, normal_meta, normal, tumor_meta, tumor, interval ->
```
pinin4fjords:
```suggestion
                          .map { grouping_key, normal, tumor, interval ->
```

docs/side_quests/splitting_and_grouping.md
```diff
@@ -925,17 +1018,17 @@ We can reuse the `subMap` method from before to isolate our `id` and `interval`
...
                               [
                                   grouping_key + [interval: interval],
                                   normal,
                                   tumor
                               ]
                           }
 
-    ch_grouped_samples = ch_combined_samples.map { grouping_key, normal, tumor ->
+    ch_grouped_samples = ch_combined_samples.map { grouping_key, normal_meta, normal, tumor_meta, tumor ->
```
pinin4fjords:
```suggestion
    ch_grouped_samples = ch_combined_samples.map { grouping_key, normal, tumor ->
```

docs/side_quests/splitting_and_grouping.md
pinin4fjords: we don't have these metas any more, after section 3.6

docs/side_quests/splitting_and_grouping.md
```diff
@@ -995,27 +1088,29 @@ Let's now group the samples by this new grouping element, using the [`groupTuple
...
-                                  normal,
-                                  tumor
-                              ]
-                          }
-                          .groupTuple()
-                          .view()
-    ```
+    ```groovy title="main.nf" linenums="39" hl_lines="8"
+
+ch_grouped_samples = ch_combined_samples.map { grouping_key, normal_meta, normal, tumor_meta, tumor ->
```
pinin4fjords:
```suggestion
ch_grouped_samples = ch_combined_samples.map { grouping_key, normal, tumor ->
```

docs/side_quests/splitting_and_grouping.md
```diff
@@ -600,11 +639,14 @@ To do so, first we define the closure as a new variable:
...
+    ```groovy title="main.nf" linenums="2" hl_lines="7"
     ch_samplesheet = Channel.fromPath("./data/samplesheet.csv")
-                        .splitCsv(header: true)
+                    .splitCsv(header: true)
+                    .map{ row ->
+                        [[id:row.id, repeat:row.repeat, type:row.type], row.bam]
+                    }
 
-    getSampleIdAndReplicate = { sample -> [ sample.subMap(['id', 'repeat']), sample ] }
+    getSampleIdAndReplicate = { meta, file -> [ meta.subMap(['id', 'repeat']), meta, file ] }
```
pinin4fjords:
```suggestion
    getSampleIdAndReplicate = { meta, file -> [ meta.subMap(['id', 'repeat']), meta, file(file) ] }
```

Maybe we should actually make this a path object at this point, and point to the files side-quest?

pinin4fjords (commented):
OK, review of splitting/ grouping.

Mostly fine, which makes sense (it was already there), couple of suggestions/ bug fixes

docs/side_quests/working_with_files.md
adamrtalbot: Yeah...it was the quickest route to supporting a remote file. 

Perhaps if we move it up to under `file()` it will be more obvious?

Flow:
 - Here's a file!
 - Here's the `file()` method
     - side quest, you can do a remote file!
 - Back to local, here's `Channel.fromPath()`
 - etc.

pinin4fjords: Yep, exactly my thinking when I suggested putting remove files earlier rather than as a bolt-on.

docs/side_quests/metadata.md
FriederikeHanssen: fine by me. I think this may have been a grammarly suggestion, so either is fine for me

docs/side_quests/metadata.md
FriederikeHanssen: maybe rather we start with `results` only and then show in the next step that the meta map can be used here as well?

docs/side_quests/metadata.md
FriederikeHanssen: Should we add something about the balance of making code more shareable vs more explicit?

docs/side_quests/metadata.md
FriederikeHanssen: 9e27b10c

FriederikeHanssen: only addressed in the main text not in the code.

docs/side_quests/metadata.md
FriederikeHanssen: 9e27b10c

docs/side_quests/metadata.md
pinin4fjords: sounds good

docs/side_quests/metadata.md
pinin4fjords: I'd say this is more explicit AND more sharable?

docs/side_quests/metadata.md
FriederikeHanssen: 20321299

docs/side_quests/metadata.md
FriederikeHanssen: 045575c6

docs/side_quests/metadata.md
FriederikeHanssen: 045575c6

docs/side_quests/metadata.md
FriederikeHanssen: 8030efbe

docs/side_quests/metadata.md
FriederikeHanssen: 7ad45c12

docs/side_quests/metadata.md
FriederikeHanssen: 7ad45c12

docs/side_quests/metadata.md
FriederikeHanssen: 3359be91

docs/side_quests/metadata.md
FriederikeHanssen: 51d9373d

docs/side_quests/metadata.md
FriederikeHanssen: I think I am being dense , that's opposite of what we are saying here: https://github.com/nextflow-io/training/pull/644#discussion_r2256718447

docs/side_quests/metadata.md
FriederikeHanssen: what I meant to say was something like: params typically applied to all files, not tailored to a specific file. Using meta data as shown here allows us to fine tune things per file.

I'll try rephrasing

FriederikeHanssen:
```suggestion
This approach differs from using pipeline parameters (`params`), which generally apply the same configuration to all files in your workflow. By leveraging metadata applied to each item in a channel, you can fine-tune process behavior on a per-file basis.
```

FriederikeHanssen: @pinin4fjords what do you think of this instead?

docs/side_quests/metadata.md
FriederikeHanssen: I am still feeling conflicted about this one. It's one of the teaching points in splitting and grouping and this module is already pretty dense. Do you have an idea on how to keep it short and sweet?  Or would as a first iteration a link to splitting and grouping help?

docs/side_quests/metadata.md
pinin4fjords: Sounds good, but maybe tweak to "By leveraging metadata **applied to each item in a channel**" to be super-clear?

docs/side_quests/metadata.md
pinin4fjords: For me that's a different situation, but maybe we do need to clean it up a bit.

Bad = allowing arbitrary channel structure (some of which a process won't care about) to dictate the process interface (which is what we're mitigating against in that comment.

Good = being explicit and having clearly named and separate inputs for the things the process **does** explicitly care about

Am I making sense?

docs/side_quests/metadata.md
pinin4fjords: Yeah, maybe even just something like:

```
Another compelling reason to use meta maps in this way is that they make it easier to associate related results that share the same identifiers. In modern Nextflow, it's considered best practice to keep different types of results in separate channels until you actually need to combine them. However, at some point, you'll inevitably reach a stage in your workflow where combining those results becomes necessary.

As you learned in Hello Nextflow, you can't rely on the order of items in channels to match results across them. Instead, you must use keys to align data correctly - and meta maps provide an ideal structure for this purpose. We explore this use case in detail in [INSERT LINK HERE].
```

FriederikeHanssen: added in c7240d3a

docs/side_quests/metadata.md
FriederikeHanssen: 361df8e4

docs/side_quests/metadata.md
FriederikeHanssen: ah yes perfectly. Added another sentence after your clarifying paragraph to sum this up: e65e4539

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1361 @@
...
+
+=== "Before"
+
+    ```groovy title="file_operations.nf" linenums="4" hl_lines="3"
+    ch_fastq.map { myFile ->
+        def (sample, replicate, type, readNum) = myFile.simpleName.tokenize('_')
+        [ sample, replicate, type, readNum.readNum, myFile ]
+    }
+    ```
+
```
pinin4fjords, line 850:
```suggestion

Notice that we're simplifying a couple of the meta data items as we go (e.g. `readNum.replace('rep', '')`).

```

docs/side_quests/working_with_files.md
```diff
@@ -0,0 +1,1361 @@
...
+```console title="Simplified Metadata Output"
+ N E X T F L O W   ~  version 24.10.4
+
+Launching `file_operations.nf` [reverent_volta] DSL2 - revision: b3aac71fea
+
+[[sample:sampleA, replicate:1, type:normal, readNum:2], /workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R2_001.fastq.gz]
+[[sample:sampleA, replicate:1, type:normal, readNum:1], /workspaces/training/side-quests/working_with_files/data/sampleA_rep1_normal_R1_001.fastq.gz]
+```
+
+Nice! we have removed the "rep" from the replicate string.
```
pinin4fjords:
```suggestion
```

docs/side_quests/working_with_files.md
pinin4fjords: implemented

PR #652, opened 2025-08-14
Add AI-assisted Groovy essentials

pinin4fjords:
Not convinced this is up to much yet, but might provide a starting point


docs/side_quests/groovy_essentials.md
```diff
@@ -0,0 +1,1725 @@
...
+    ```
+
+This shows realistic Nextflow patterns:
+
+- **Separate processes** for different strategies rather than dynamic generation
+- **When conditions** to control which processes run for which samples
+- **Mix operator** to combine results from different conditional processes
+- **Process parameterization** using metadata in script blocks
+
+### 3.3. Channel-based Workflow Routing
```
FriederikeHanssen: I feel like this could live in a non groovy training too

side-quests/solutions/essential_scripting_patterns/main.nf
```diff
@@ -0,0 +1,69 @@
...
+
+    return [sample_meta + file_meta + [priority: priority], fastq_path]
+}
+
+workflow {
+    validateInputs()
+
+    ch_samples = Channel.fromPath(params.input)
+        .splitCsv(header: true)
+        .map{ row -> separateMetadata(row) }
```
mribeirodantas, line 52: Should we mention that there is a more Groovy/concise way for this that is simply:
> .map(separateMetadata)

?

I tend to like explicit more than implicit, but not sure if in this case it'd be worth it mentioning the concise version.

docs/side_quests/groovy_essentials.md
```diff
@@ -0,0 +1,2207 @@
...
+
+```groovy title="main.nf" linenums="1"
+include { FASTP } from './modules/fastp.nf'
+```
+
+Then modify the `workflow` block to connect the `ch_samples` channel to the `FASTP` process:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="25" hl_lines="7"
```
mribeirodantas:
```suggestion
    ```groovy title="modules/fastp.nf" linenums="25" hl_lines="7"
```

docs/side_quests/groovy_essentials.md
```diff
@@ -0,0 +1,2207 @@
...
+            .splitCsv(header: true)
+            .map{ row -> separateMetadata(row) }
+
+        ch_fastp = FASTP(ch_samples)
+    }
+    ```
+
+=== "Before"
+
+    ```groovy title="main.nf" linenums="25" hl_lines="6"
```
mribeirodantas:
```suggestion
    ```groovy title="modules/fastp.nf" linenums="25" hl_lines="6"
```

side-quests/solutions/essential_scripting_patterns/main.nf
pinin4fjords: Thanks @mribeirodantas, but to make that work it would have to be defined as a closure, within the `workflow` block, rather than as a function definition as I've done here. 

I asked @bentsherman about using closures vs functions in such situations, and he advised that functions should be the convention.

docs/side_quests/groovy_essentials.md
pinin4fjords: We're editing the main.nf, not the fastp.nf, so this is not a correct change.

docs/side_quests/groovy_essentials.md
pinin4fjords: We're editing the main.nf, not the fastp.nf, so this is not a correct change.

bentsherman:
I think this content is awesome, but I have a higher-level objection to the use of "Groovy" at all.

The [Nextflow language specification](https://nextflow.io/docs/latest/reference/syntax.html) was created in part to make it easier to talk about Nextflow as a standalone language. So rather than saying "Nextflow is based on Groovy but we promise it's not that bad!" we can instead say "Nextflow is a programming language that runs on the JVM".

It also gets confusing when talking about things like "Nextflow operators" vs "Groovy functions"... well, Nextflow operators are Groovy functions! They have the same syntax and (mostly) the same semantics

And it will get more confusing in 25.10 when it becomes possible to write code like this:

```groovy
def sample_ids: List<String> = ['sample_001', 'sample_002', 'sample_003']
```

Which this tutorial might call "pure Groovy code", but isn't really pure Groovy because it isn't valid Groovy syntax. The distinction between Nextflow and Groovy will become less and less useful over time as the languages diverge futher

I think the distinction you are usually getting at with "Nextflow constructs" vs "Groovy constructs" is dataflow logic in the workflow vs "regular" logic (procedural logic?) in places like an operator closure. So I would describe it in those terms instead of Nextflow vs Groovy

Similarly, on the section about `collect` operator vs Groovy's `collect` method, well... this method is in the Nextflow [standard library](https://nextflow.io/docs/latest/reference/stdlib-types.html#iterable-e)! So you don't even need to talk about Groovy, you can just say it's the `collect` method of the `Iterable` type vs the `Channel` type

In fact, I think you should be able to link any piece of syntax or standard library to the Nextflow reference docs -- and if you can't find it, tell me and I will add it. I want to have the Nextflow language fully specified within our docs so that we don't have to mention Groovy at all. I mean it

docs/side_quests/groovy_essentials.md
```diff
@@ -0,0 +1,2283 @@
...
+    def log_file = file("${workflow.launchDir}/pipeline_summary.txt")
+    log_file.text = summary
+}
+```
+
+### 8.2. Other Useful Event Handlers
+
+Besides `onComplete`, there are other event handlers you can use:
+
+**`onStart`** - Runs when the workflow begins:
```
bentsherman: There is no such thing as the `onStart` workflow handler

pinin4fjords: Ooops, AI stuff I neglected to strip- thanks Ben

pinin4fjords:
Thanks for the feedback @bentsherman, really useful. I'll re-pitch and get you to re-review if you get time.

Edit: done, with Claude's help.

I've used 'scripting' as the counterpoint to 'dataflow', because we're not necessarily talking to a crowd to whom 'procedural' makes enough sense. See what you think.

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
+# Essential Nextflow Scripting Patterns
+
+Nextflow is a programming language that runs on the Java Virtual Machine. While Nextflow was originally built on Groovy and shares much of its syntax, the [Nextflow language specification](https://nextflow.io/docs/latest/reference/index.html) defines Nextflow as a standalone language—making it easier to understand Nextflow on its own terms rather than as "Groovy with extensions."
```
bentsherman:
```suggestion
Nextflow is a programming language that runs on the Java Virtual Machine. While Nextflow is built on [Groovy](http://groovy-lang.org/) and shares much of its syntax, Nextflow is more than just "Groovy with extensions" -- it is a standalone language with a fully-specified [syntax](https://nextflow.io/docs/latest/reference/syntax.html) and [standard library](https://nextflow.io/docs/latest/reference/stdlib.html).
```

Link was incorrect but I made a few other suggestions

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+## 0. Warmup
+
+### 0.1. Prerequisites
+
+Before taking on this side quest you should:
+
+- Complete the [Hello Nextflow](../hello_nextflow/README.md) tutorial or have equivalent experience
+- Understand basic Nextflow concepts (processes, channels, workflows)
+- Have basic familiarity with common programming constructs used in Groovy syntax (variables, maps, lists)
```
bentsherman:
```suggestion
- Have basic familiarity with common programming constructs (variables, maps, lists)
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+This tutorial will explain Groovy concepts as we encounter them, so you don't need extensive prior Groovy knowledge. We'll start with fundamental concepts and build up to advanced patterns.
```
bentsherman:
```suggestion
This tutorial will explain programming concepts as we encounter them, so you don't need extensive programming experience. We'll start with fundamental concepts and build up to advanced patterns.
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+When writing Nextflow workflows, it's important to distinguish between **dataflow** (how data moves through channels and processes) and **scripting** (the code that manipulates data and makes decisions). Let's build a workflow demonstrating how they work together.
+
+#### Step 1: Basic Nextflow Workflow
+
+Start with a simple workflow that just reads the CSV file (we've already done this for you in `main.nf`):
+
+```groovy title="main.nf" linenums="1"
+workflow {
+    ch_samples = Channel.fromPath("./data/samples.csv")
```
bentsherman:
```suggestion
    ch_samples = channel.fromPath("./data/samples.csv")
```

Trying to standardize around using the `channel` namespace to access channel factories. Should be easy to search & replace across the board

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+    ```groovy title="main.nf" linenums="2" hl_lines="3"
+        ch_samples = Channel.fromPath("./data/samples.csv")
+            .splitCsv(header: true)
+            .view()
+    ```
+
+This is our first **closure**—an anonymous function you can pass as an argument (similar to lambdas in Python or arrow functions in JavaScript). Closures are essential for working with Nextflow operators.
+
+The closure `{ row -> return row }` takes a parameter `row` (could be any name: `item`, `sample`, etc.). You can also use the implicit variable `it` instead: `.map { return it }`, though naming parameters improves clarity.
```
bentsherman:
```suggestion
The closure `{ row -> return row }` takes a parameter `row` (could be any name: `item`, `sample`, etc.).
```

Not a hard suggestion, because it might still be useful to mention this since it is still commonly used out in the wild, but implicit `it` is basically deprecated in the Nextflow language. One less variant to worry about

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+```bash title="Test map operator"
+nextflow run main.nf
+```
+
+You'll see the same output as before, because we're simply returning the input unchanged. This confirms that the map operator is working correctly. Now let's start transforming the data.
+
+#### Step 3: Creating a Map Data Structure
+
+Now we're going to write **scripting** inside our closure to transform each row of data. This is where we process individual data items rather than orchestrating data flow.
```
bentsherman:
```suggestion
Now we're going to write **scripting** logic inside our closure to transform each row of data. This is where we process individual data items rather than orchestrating data flow.
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+!!! tip "Map Operations Summary"
+
+    - **Add keys**: `map1 + [new_key: value]` - Creates new map with additional keys
+    - **Extract keys**: `map1.subMap(['key1', 'key2'])` - Creates new map with only specified keys
+    - **Both operations create new maps** - Original maps remain unchanged
+
+#### Step 5: Combining Maps and Returning Results
+
+So far, we've only been returning what Nextflow community calls the 'meta map', and we've been ignoring the files those metadata relate to. But if you're writing Nextflow workflows, you probably want to do something with those files.
```
bentsherman:
```suggestion
So far, we've only been returning what the Nextflow community calls the 'meta map', and we've been ignoring the files those metadata relate to. But if you're writing Nextflow workflows, you probably want to do something with those files.
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+            .map { row ->
+                def sample_meta = [
+                    id: row.sample_id.toLowerCase(),
+                    organism: row.organism,
+                    tissue: row.tissue_type.replaceAll('_', ' ').toLowerCase(),
+                    depth: row.sequencing_depth.toInteger(),
+                    quality: row.quality_score.toDouble()
+                ]
+                def priority = sample_meta.quality > 40 ? 'high' : 'normal'
+                return [sample_meta + [priority: priority], file(row.file_path) ]
```
bentsherman:
```suggestion
                return tuple( sample_meta + [priority: priority], file(row.file_path) )
```

Always use `tuple()` instead of a list when you want to create a tuple -- you'll thank me later when we have type checking

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+This `[meta, file]` tuple structure is a common pattern in Nextflow for passing both metadata and associated files to processes.
+
+!!! note
+
+    **Maps and Metadata**: Maps are fundamental to working with metadata in Nextflow. For a more detailed explanation of working with metadata maps, see the [Working with metadata](./metadata.md) side quest.
+
+Our workflow demonstrates the core pattern: **dataflow operations** (`workflow`, `Channel.fromPath()`, `.splitCsv()`, `.map()`, `.view()`) orchestrate how data moves through the pipeline, while **scripting** (maps `[key: value]`, string methods, type conversions, ternary operators) inside the `.map()` closure handles the transformation of individual data items.
+
+### 1.2. Understanding Different Types: Channel vs Iterable
```
bentsherman: It might be easier to just talk about List here instead of Iterable. The whole "iterable is a trait implemented by List / Set / Bag" is probably not a can of worms we want to get into here

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+    ```groovy title="modules/generate_report.nf" linenums="10"
+        script:
+        """
+        echo "Processing ${reads}" > ${meta.id}_report.txt
+        echo "Sample: ${meta.id}" >> ${meta.id}_report.txt
+        """
+    ```
+
+If you run this, you'll notice an error or unexpected behavior - Nextflow tries to interpret `$(hostname)` as a Groovy variable that doesn't exist:
```
bentsherman:
```suggestion
If you run this, you'll notice an error or unexpected behavior - Nextflow tries to interpret `$(hostname)` as a Nextflow variable that doesn't exist:
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+This makes your workflows both more efficient (not over-allocating) and more robust (automatic retry with more resources).
+
+---
+
+## 5. Conditional Logic and Process Control
+
+Previously, we used `.map()` with scripting to transform channel data. Now we'll use conditional logic to control which processes execute based on data—essential for flexible workflows adapting to different sample types.
+
+Nextflow's [flow control operators](https://www.nextflow.io/docs/latest/reference/operator.html) take closures evaluated at runtime, enabling conditional logic to drive workflow decisions based on channel content.
```
bentsherman:
```suggestion
Nextflow's [dataflow operators](https://www.nextflow.io/docs/latest/reference/operator.html) take closures evaluated at runtime, enabling conditional logic to drive workflow decisions based on channel content.
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+ N E X T F L O W   ~  version 25.04.3
+
+Launching `main.nf` [trusting_torvalds] DSL2 - revision: b56fbfbce2
+
+ERROR ~ Cannot invoke method toUpperCase() on null object
+
+ -- Check script 'main.nf' at line: 13 or see '.nextflow.log' file for more details
+```
+
+The problem is that `row.run_id` returns `null` because the `run_id` column doesn't exist in our CSV. When we try to call `.toUpperCase()` on `null`, it crashes. This is where Groovy's safe navigation operator saves the day.
```
bentsherman:
```suggestion
The problem is that `row.run_id` returns `null` because the `run_id` column doesn't exist in our CSV. When we try to call `.toUpperCase()` on `null`, it crashes. This is where the safe navigation operator saves the day.
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+```bash
+nextflow run main.nf
+```
+
+No crash! The workflow now handles the missing field gracefully. When `row.run_id` is `null`, the `?.` operator prevents the `.toUpperCase()` call, and `run_id` becomes `null` instead of causing an exception.
+
+### 6.3. Elvis Operator (`?:`) for Defaults
+
+The Elvis operator (`?:`) provides default values when the left side is `null` (or empty, in Groovy's "truth" evaluation). It's named after Elvis Presley because `?:` looks like his famous hair and eyes when viewed sideways!
```
bentsherman:
```suggestion
The Elvis operator (`?:`) provides default values when the left side is "falsy" (as explained previously). It's named after Elvis Presley because `?:` looks like his famous hair and eyes when viewed sideways!
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+- **Elvis operator (`?:`)**: Provides defaults - `value ?: 'default'`
+- **Combining**: `value?.method() ?: 'default'` is the common pattern
+
+These operators make workflows resilient to incomplete data - essential for real-world work.
+
+---
+
+## 7. Validation with `error()` and `log.warn`
+
+Sometimes you need to stop the workflow immediately if input parameters are invalid. While `error()` and `log.warn` are Nextflow-provided functions, the **validation logic itself uses standard programming constructs**—conditionals (`if`, `!`), boolean logic, and methods like `.exists()`. Let's add validation to our workflow.
```
bentsherman:
```suggestion
Sometimes you need to stop the workflow immediately if input parameters are invalid. In Nextflow, you can use built-in functions like `error()` and `log.warn`, as well as standard programming constructs like `if` statements and boolean logic, to implement validation logic. Let's add validation to our workflow.
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+- **Early validation**: Check inputs before processing to fail fast with helpful errors
+- **Validation functions**: Create reusable validation logic that can be called at workflow start
+
+Proper validation makes workflows more robust and user-friendly by catching problems early with clear error messages.
+
+---
+
+## 8. Configuration and Workflow Event Handlers
+
+Up until now, we've been writing code in our workflow scripts and process definitions. But there's one more important place where you can add logic: workflow event handlers in your `nextflow.config` file (or other places you write configuration).
```
bentsherman: Workflow handlers in the config are deprecated: https://nextflow.io/docs/latest/config.html#workflow-handlers

Our docs recommend to use them only in the script, and for use cases that require a config handler, implement as a plugin instead. So I think you could just refactor this section to show the workflow handler in a script instead

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+
+    println summary
+
+    // Write to a log file
+    def log_file = new File("${workflow.launchDir}/pipeline_summary.txt")
+    log_file.text = summary
+}
+```
+
+### 8.2. Other Useful Event Handlers
```
bentsherman: Seems like this section could just be "The onError handler" since that is the only other handler

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+    Completed: ${workflow.complete}
+    Duration : ${workflow.duration}
+    Success  : ${workflow.success}
+    Command  : ${workflow.commandLine}
+    """
+
+    println summary
+
+    // Write to a log file
+    def log_file = new File("${workflow.launchDir}/pipeline_summary.txt")
```
bentsherman: Try to avoid the use of `new` if there is a more "standard" alternative. In this case you should be able to use the `file` function, although maybe it doesn't work in the config

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+  ```groovy title="Error handling patterns"
+  try {
+      def errors = validateSample(sample)
+      if (errors) throw new RuntimeException("Invalid: ${errors.join(', ')}")
+  } catch (Exception e) {
+      println "Error: ${e.message}"
+  }
+  ```
+
+- **Essential Groovy Operators**
```
bentsherman:
```suggestion
- **Essential Operators**
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+  // Filter, group, and organize data
+  def high_quality = samples.findAll { it.quality > 40 }
+  def by_organism = samples.groupBy { it.organism }
+  def file_names = files*.getName()  // Spread operator
+  def all_files = nested_lists.flatten()
+  ```
+
+## Resources
+
+- [Nextflow Language Reference](https://nextflow.io/docs/latest/reference/index.html)
```
bentsherman:
```suggestion
- [Nextflow Language Reference](https://nextflow.io/docs/latest/reference/syntax.html)
```

docs/side_quests/essential_scripting_patterns.md
```diff
@@ -0,0 +1,2290 @@
...
+- [Nextflow Operators](https://www.nextflow.io/docs/latest/operator.html)
+- [Nextflow Script Syntax](https://www.nextflow.io/docs/latest/script.html)
+- [Groovy Documentation](http://groovy-lang.org/documentation.html) (for deeper understanding of underlying language features)
```
bentsherman:
```suggestion
- [Nextflow Standard Library](https://nextflow.io/docs/latest/reference/stdlib.html)
```

docs/side_quests/essential_scripting_patterns.md
pinin4fjords: Yeah, this was a fix because file() didn't work, but maybe it will resolve if we move it.

PR #683, opened 2025-10-21
Add Python 3.13 to devcontainer to support latest nf-core tools

pinin4fjords:
## Problem

The devcontainer configuration was unable to install the latest version of nf-core tools. When `uv tool install nf-core` ran, it was using Ubuntu 22.04's system Python 3.10 to create the isolated environment, which limited it to installing only versions of nf-core compatible with Python 3.10. If the latest nf-core requires Python 3.13, an older version would be installed instead.

## Solution

This PR adds the official `ghcr.io/devcontainers/features/python:1` feature with version 3.13 to both the `codespaces-dev` and `local-dev` devcontainer configurations. By installing Python 3.13 early in the feature installation order (before `uv`), we ensure that when `uv tool install nf-core` runs, it will:

1. Find Python 3.13 as the system Python
2. Create the isolated environment with Python 3.13
3. Install the latest version of nf-core that requires Python 3.13

## Changes

- Added Python 3.13 feature to [codespaces-dev/devcontainer.json](.devcontainer/codespaces-dev/devcontainer.json)
- Added Python 3.13 feature to [local-dev/devcontainer.json](.devcontainer/local-dev/devcontainer.json)
- Updated [uv-tools feature](.devcontainer/local-features/uv-tools/devcontainer-feature.json) to declare dependency on Python feature installation for proper ordering

## Design Rationale

This approach aligns with the native installation philosophy introduced in #609:
- Uses native Python installation (via devcontainer feature) rather than conda
- Keeps conda available exclusively for pipeline execution within training materials
- Avoids hardcoding paths or pinning specific tool versions
- Ensures `uv` automatically gets the latest compatible nf-core version

## Testing

The changes should be tested by:
1. Rebuilding the devcontainer
2. Verifying Python version: `python --version` (should show 3.13.x)
3. Checking nf-core installation: `nf-core --version` (should show latest version)
4. Confirming uv is using Python 3.13: `/root/.local/share/uv/tools/nf-core/bin/python --version`

Related to #609

### DIFF
--- a/.devcontainer/local-features/uv-tools/devcontainer-feature.json
+++ b/.devcontainer/local-features/uv-tools/devcontainer-feature.json
@@ -2,5 +2,8 @@
     "id": "uv-tools",
     "name": "uv tools",
     "description": "Install required python cli tools using uv",
-    "installsAfter": ["ghcr.io/va-h/devcontainers-features/uv", "ghcr.io/devcontainers/features/python"]
+    "installsAfter": [
+        "ghcr.io/va-h/devcontainers-features/uv",
+        "ghcr.io/devcontainers/features/python"
+    ]
 }


pinin4fjords:
Note: build logs now show correct version of nf-core tools, so I'm pretty sure this has worked

pinin4fjords:
Thanks @kenibrewer !

PR #686, opened 2025-10-28
Replace "Channel" with the lowercase "channel"

robsyme:
The uppercase version no longer provides autocomplete for the Nextflow language server.


mribeirodantas (approved):

mribeirodantas:
It hurts me to use _channel_ instead of _Channel_, but c'est la vie 😆

PR #687, opened 2025-10-28
bump MQC container

matthdsm:
This should fix the Mac arm compatibility issues some users are having.

Cheers


maxulysse:
LGTM, but I cannot test as I don't have a mac

vdauwera:
There were some legacy errors in the the header numbering (my fault from a while back), fixing now.

PR #692, opened 2025-11-04
NF4Science course template

vdauwera:
This is meant to be a course template to make our NF4Science mini-courses consistent going forward. It's based mainly on the RNAseq lesson plan, fleshing that out further. The general format is aimed at demonstrating the concepts and mechanisms presented in Hello Nextflow, but in the context of a scientific use case.

The base plan that should be used for all NF4S courses is as follows.

- [x] Index
- [x] Orientation
- [x] Method overview (run commands manually in containers)
- [ ] Single-sample implementation (show basic usage like in Hello BUT start with a samplesheet input and modules)
- [ ] Multi-sample implementation (show use of `collect()` and other relevant operators, any domain-specific quirks etc)

I envision optional add-ons such as:

- [ ] Configuration / resource optimization
- [ ] Make the workflow composable
- [ ] Handle metadata propagation
- [ ] Testing with nf-test

(Optimal order is TBD)

### DIFF
--- a/docs/nf4_science/course-template/02_single-sample.md
+++ b/docs/nf4_science/course-template/02_single-sample.md
@@ -1,0 +1,557 @@
+# Part 2: Single-sample implementation
+
+In this part of the course, we're going to write the simplest possible workflow that wraps all the commands we ran in Part 1 to automate running them.
+
+We'll do this in three stages:
+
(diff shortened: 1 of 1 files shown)


vdauwera:
Rebooting this from the updated/standardized Genomics and RNAseq courses, using Claude.

PR #699, opened 2025-11-18
Fixes map_demo.nf location according to training instructions

mribeirodantas:
Created from https://community.seqera.io/t/metadata-side-quest/2454


mribeirodantas:
Thanks, @FriederikeHanssen !

PR #704, opened 2025-11-21
Extended validation to cover two more fields in the CSV

vdauwera:
This is to match an upcoming change in the Hello Nextflow greetings file, and provides opportunity to see other types of validation rules in action

### DIFF
--- a/docs/hello_nf-core/04_make_module.md
+++ b/docs/hello_nf-core/04_make_module.md
@@ -833,7 +833,7 @@ Each file serves a specific purpose:
 
 !!! tip "Learn more about testing"
 
-    The generated test file uses nf-test, a testing framework for Nextflow pipelines and modules. To learn how to write and run these tests, see the [nf-test side quest](../../side_quests/nf_test/).
+    The generated test file uses nf-test, a testing framework for Nextflow pipelines and modules. To learn how to write and run these tests, see the [nf-test side quest](../side_quests/nf_test.md).
 
 The generated `main.nf` includes all the patterns you just learned, plus some additional features:
 


vdauwera:
I think this should all work but haven't retested the validation piece yet

vdauwera:
Finished testing, this all works in my hands

docs/hello_nf-core/03_use_module.md
mribeirodantas: It did not prompt anything here :|

Image

docs/hello_nf-core/04_make_module.md
mribeirodantas, line 183: I get:

```

!! Only displaying parameters that differ from the pipeline defaults !!
------------------------------------------------------
[-        ] CORE_HELLO:HELLO:sayHello       -
[-        ] CORE_HELLO:HELLO:convertToUpper -
Process `CORE_HELLO:HELLO:cowpy` declares 2 input channels but 1 were specified

 -- Check script 'workflows/hello.nf' at line: 44 or see '.nextflow.log' file for more details
```

docs/hello_nf-core/03_use_module.md
vdauwera: Hmm weird, for me one time it did, and the next it didn't. I adapted the wording to be conditional.

docs/hello_nf-core/04_make_module.md
vdauwera: Are you sure you followed the instructions correctly? The two inputs to the cowpy process get set up in Part 2, so if the call worked at the end of Part 2 and in Part 3, there's no reason it wouldn't continue to work in Part 4 unless you somehow deleted the second input by mistake when updating the call to cowpy. I tested this starting from the Part 3 solution, and everything worked as expected.

PR #727, opened 2025-11-25
Improve the layout of the intro and outro materials for Side Quests

vdauwera:
Toying with some ideas to improve the layout of the Side Quests.

- Distinguish and highlight **Learning goals** and **Prerequisites** with subheadings in the intro part
-  Rename the warmup section (Get started) and give it more structure
    - Open the training codespace
    - Move into the project directory
    - Explore the materials
    - Scenario
    - Readiness checklist
- Summary: associate the code examples with the "what you've learned" recaps and streamline the AI-generated text. Some of it is repetitive.
- Add/rework "What's next?" sections

### DIFF
--- a/.github/workflows/nextflow-lint-comment.yml
+++ b/.github/workflows/nextflow-lint-comment.yml
@@ -1,0 +1,48 @@
+name: Nextflow Lint Comment
+
+on:
+  workflow_run:
+    workflows: ["Nextflow Lint"]
+    types:
+      - completed
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  comment:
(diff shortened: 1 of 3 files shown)


pinin4fjords (approved):
Nice work!

The standardized structure for intro/outro sections is a clear improvement - having consistent "Learning goals", "Prerequisites", and "Get started" sections across all side quests will help learners know what to expect.

Adding code snippets to the Summary sections is particularly useful - makes them work as quick reference rather than just recap text.

One small note: "What's next?" uses `##` while the intro subsections use `###`, so the ending ends up more prominent than the beginning. Not a blocker, just flagging in case it wasn't intentional.

vdauwera:
Thanks @pinin4fjords ! Yep that ## was intentional, it's so the what's next stands out vs the summary.

PR #742, opened 2025-12-03
[EXAMPLE] hello-nextflow: new Nextflow syntax (full)

ewels:
I didn't really mean to do this, I was just sort of curious and prodding around with @claude, then one thing led to another, and here we are.. 🤷🏻 

This PR includes \*all\* upcoming syntax changes. Primarily:

* Workflow inputs (params) - typed
* Workflow outputs
* Typed process inputs

It is meant as a demonstrator of what kind of changes will be needed for the training material - I'm not expecting that we merge as-is. It's possible that we can slice off some of this work and repackage in another PR to kickstart the update process.


docs/hello_nextflow/00_orientation.md
```diff
@@ -17,6 +17,41 @@
...
+    nextflow -version
+    ```
+
+### Enable the v2 syntax parser
+
+The typed syntax features used in this course require enabling Nextflow's v2 syntax parser.
+Set the following environment variable before running any workflows:
+
+```bash
+export NXF_SYNTAX_PARSER=v2
```
vdauwera, line 38: Are we going to set this by default in the training environment? If so, can we make this a much shorter note to tell people running in their own environment to check requirements for local installation, and have the details live there? I really don't want to add a half page of warnings about previews and syntax requirements in a beginner training intro.

ewels: We can do, but that'll tell Nextflow to use the new syntax parser for _all_ the training courses. Which we don't want yet, until they're all updated.

This is where timelines matter. This will become the default in 2026, so won't be needed from that point onwards. So if we're fast at updating all the courses, then yes - let's stick it into the dev environment. If not, then I think we need to tell people to do it themselves.

Probably the most user-friendly way to do this is to have an interactive script that runs on codespaces init that prompts the user to ask which course they're doing, then sets some env variables accordingly.

Or, _\*controvercy alert\*_, we split up the courses into separate repos and/or containers and have one per course.

vdauwera: Hahaha I hate both of those options. I don't want to have major differences in requirements for different courses _except_ for specific super advanced topics, and they all have to work within a single environment.

So we just update syntax for everything in one go. It's the most sane option for end-users.

vdauwera:
> I do not think we should include preview features in the training material

Completely agree. We've discussed this over slack and are making a plan to do the updates for released features now (use strict syntax across all courses) and another batch mid-year once the preview features are released.

ewels:
> I do not think we should include preview features in the training material

Yeah apologies - I opened the PR in tandem with an ongoing Slack thread so the context there was clear, but not when seeing the PR stand alone. I've updated the PR description and status to make this clear.

pditommaso:
Awesome, thanks 🙏

docs/hello_nextflow/01_hello_world.md
```diff
@@ -104,7 +104,7 @@ Let's open the `hello-world.nf` script in the editor pane.
 process sayHello {
 
     output:
-        path 'output.txt'
+    file 'output.txt'
```
bentsherman, line 107: Minor point since typed processes shouldn't be used here until they are stable, but the "canonical" syntax here should use parentheses because it is just a function call returning a value, not a type qualifier like in the old process syntax

ewels: Worth updating [the docs](https://www.nextflow.io/docs/latest/process-typed.html#file-inputs) in that case?

hello-nextflow/hello-config.nf
```diff
@@ -1,11 +1,15 @@
 #!/usr/bin/env nextflow
 
+nextflow.preview.types = true
+
```
bentsherman, line 4: Not needed here. Did the language server add this? It should only add this flag if there were processes that were auto-converted

ewels: No this was me / claude being over-zealous in adding it. I didn't do any language server auto-correction here.

docs/hello_nextflow/01_hello_world.md
bentsherman: I think [the docs](https://www.nextflow.io/docs/latest/process-typed.html#outputs) are pretty clear

ewels:
Ok I pulled out the parts of this PR that are for inputs + outputs + strict syntax and put them into a separate PR: https://github.com/nextflow-io/training/pull/743

Can leave this one here for future reference if we want, or just close it out (can always refer to it in closed state).

vdauwera:
> Ok I pulled out the parts of this PR that are for inputs + outputs + strict syntax and put them into a separate PR: #743
> 
> Can leave this one here for future reference if we want, or just close it out (can always refer to it in closed state).

Let's close it out to avoid confusion

docs/hello_nextflow/01_hello_world.md
ewels: Sorry, too many types!

PR #743, opened 2025-12-03
Hello nextflow: New workflow inputs + outputs + strict syntax

ewels:
Rehash of https://github.com/nextflow-io/training/pull/742 but without the typed process inputs (which are still in beta).

Should be a more realistic start at updating the materials for the near future. We can approach typed processes + record types in 2026.

Also ran the auto-formatter on the files, which works now that strict syntax is enforced. I used the default settings - no Harshil alignment etc.

### DIFF
--- a/.github/workflows/nextflow-lint-comment.yml
+++ b/.github/workflows/nextflow-lint-comment.yml
@@ -37,7 +37,7 @@ jobs:
         with:
           issue-number: ${{ steps.pr-info.outputs.pr_number }}
           comment-author: "github-actions[bot]"
-          body-includes: "Nextflow Lint Results"
+          body-includes: "Nextflow linting complete!"
 
       - name: Create or update PR comment
         uses: peter-evans/create-or-update-comment@v4


hello-nextflow/hello-channels.nf
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env nextflow
 
+/*
+ * Pipeline parameters
+ */
+params {
+    greeting: String = 'Holà mundo!'
+}
```
bentsherman: Minor point, but regardless of whether you put them at the top or bottom, I would keep the params / workflow / output blocks together. They are essentially a unit

ewels: Ok, I reorganised the files. I also ran `nextflow lint -format` whilst I was here. Only one case of it removing a chunk of comments 😅

vdauwera:
Closing this PR in favor of https://github.com/nextflow-io/training/pull/751 which builds on the branch

PR #745, opened 2025-12-04
Fix bug with linting results comments.

ewels:
Action should update existing comments instead of posting new ones, but there was an error from it looking for some outdated text which was changed. Merging this should stop it from posting new comments each time a new commit is pushed to a PR.


ewels:
Self-merging as is very straightforward and doesn't affect anything else.

PR #748, opened 2025-12-04
Improvements to Side Quest: Metadata

vdauwera:
Based on 2025 teaching experience

Mainly improvements to instructions and consistent formatting. More detail across the board. Reordering section 3 to make more intuitive sense (see commit notes)


vdauwera:
This is ready for review. I made some important changes to section 3 to streamline and explain more clearly what was going on.

side-quests/metadata/main.nf
```diff
@@ -1,5 +1,5 @@
 workflow  {
 
-    ch_samplesheet = channel.fromPath("./data/samplesheet.csv")
+    ch_datasheet = channel.fromPath("./data/datasheet.csv")
```
pinin4fjords, line 5: I'm actually torn on this. I appreciate the better accuracy of 'datasheet', but 'samplesheet' is the nf-core convention so we might be introducing unnecessary friction for folks who move on to look at the nf-core stuff.

docs/side_quests/metadata.md
```diff
@@ -71,18 +71,20 @@ You'll find a main workflow file and a `data` directory containing a samplesheet
...
-│   └── samplesheet.csv
+│   └── datasheet.csv
 ├── main.nf
 └── nextflow.config
 ```
 
-The samplesheet list the paths to the data files and some associated metadata, organized in 3 columns:
+The workflow in the `main.nf` file is a stub that you will gradually expand into a fully functioning workflow.
+
+The datasheet list the paths to the data files and some associated metadata, organized in 3 columns:
```
pinin4fjords:
```suggestion
The datasheet lists the paths to the data files and some associated metadata, organized in 3 columns:
```

docs/side_quests/metadata.md
```diff
@@ -121,422 +123,611 @@ If you can check all the boxes, you're good to go.
...
+- `recording`
 
-    Even though it doesn't have a proper `workflow` block, Nextflow can run this as if it were a workflow:
+This is great! It makes it easy to access specific fields from each file.
+For example, we could access the file ID with `id` or the txt file path with `recording`.
 
-    ```bash title="Run map demo example"
-    nextflow run examples/map_demo.nf
-    ```
+<details>
```
pinin4fjords: I still think the [mkdocs admonition syntax](https://squidfunk.github.io/mkdocs-material/reference/admonitions/) would be easier to maintain, and we have different 'types' we can use. I'm sad that we're spreading this more complex syntax here rather than standardising on the mkdocs format elsewhere.

But you're the boss of this ;-)

docs/side_quests/metadata.md
```diff
@@ -604,254 +797,750 @@ In this section, you've learned how to :
...
+Our channel contains tuples structured as `[meta, file]`, so we can access the `file` object directly, and we can access the `character` value stored inside the meta map by referring to it as `meta.character`.
+
+In the main workflow, make the following code changes:
+
+=== "After"
+
+    ```groovy title="main.nf" linenums="34"
+        // Temporary: access the file and character
+        ch_languages.map { meta, file -> file } .view()
+        ch_languages.map { meta, file -> meta.character } .view()
```
pinin4fjords:
```suggestion
        ch_languages.map { meta, file -> file }.view{file -> "File: " + file}
        ch_languages.map { meta, file -> meta.character }.view{character -> "Character: " + character}
```

pinin4fjords: Can be hard to read multiple view outputs together without using the closure to view

docs/side_quests/metadata.md
```diff
@@ -604,254 +797,750 @@ In this section, you've learned how to :
...
+    ```groovy title="main.nf" linenums="34"
+        // Temporary: access the file and character
+        ch_languages.map { meta, file -> [file, meta.character] }
+            .view()
+    ```
+
+You see we simply copy the two map operations (minus the `.view()` statements) as the inputs to the process call.
+Just make sure you don't forget the comma between them!
+
+In case you're wondering, we can't just write a single map operation that outputs both the file and the character because that would return a tuple.
```
pinin4fjords: Actually, while this is literally true of `map` we could use multiMap here. Should we? https://www.nextflow.io/docs/latest/reference/operator.html#multimap

That returns a multi-channel object, so I _think_ the following would work:

    ```
        // Run cowpy to generate ASCII art
        COWPY(
            ch_languages.multiMap { meta, file ->
                file: file
                character: meta.character
            }
        )
    ```

But that is introducing a couple of fairly complicated concepts.

pinin4fjords (approved):
Minor comments

docs/side_quests/metadata.md
vdauwera: Oh I'm an idiot, I didn't realize we could easily choose from different types when using the ??? syntax too. My bad, will fix.

side-quests/metadata/main.nf
vdauwera: I'm a bit torn too tbh. I definitely don't want to add friction, but samplesheet just feels so omics-centric. My angle here is that if we're serious about making the training as domain-agnostic as possible, we should us terminology that is also domain-agnostic. (especially given this side quest is totally domain-agnostic)

vdauwera: I think the right thing to do is to use datasheet more widely and make sure to introduce samplesheet appropriately where it makes sense. On further reflection it feels like 'samplesheet is a special kind of datasheet' is a very easy concept and should not be a serious obstacle to our audience.

docs/side_quests/metadata.md
vdauwera: Ahhhh I was thinking it should be possible somehow, didn't think of multiMap. Thank you!

Added an optional info note to that effect. It's not too complicated, I think it's worth mentioning.

side-quests/metadata/main.nf
pinin4fjords: OK, sounds good

PR #749, opened 2025-12-26
Fixes to Hello nf-core v. 1.1

vdauwera:
Assorted improvements following 2025 Q4 season

- [x] Add note about subworkflows having a bigger role than just housekeeping in other pipelines in Part 1 
- [x] Add diagram for nested workflow structure in Part 2
- [x] Add diagram for module interfaces comparison in Part 3
- [x] Add diagram for COWPY inputs in Part 4
- [x] Add final step of testing the alt cowpy module in Part 4
- [x] Update intermediates 
- [x] Assorted minor fixes


hello-nf-core/solutions/core-hello-part4/modules/local/cowpy/main.nf
```diff
@@ -1,29 +1,28 @@
 
 
-process COWPY {
+process cowpy {
```
pinin4fjords: why are we dropping the nf-core style uppercase convention?

pinin4fjords (commented):
Mostly makes sense, like the diagrams! Feedback:

1. I find the table icon in the image for metadata problematic. For any given instance of a process they'll usually be getting a single row of that sample sheet, so I find showing a table there a bit confusing.
2. Why are we moving to lowercase process name given the nf-core convention for uppercase?

hello-nf-core/solutions/core-hello-part4/modules/local/cowpy/main.nf
vdauwera: PEBKAC, will fix

vdauwera:
> 1. I find the table icon in the image for metadata problematic. For any given instance of a process they'll usually be getting a single row of that sample sheet, so I find showing a table there a bit confusing.

Hmmm I see your point. Would it be better if we showed one row highlighted?

metamap-row

hello-nf-core/solutions/core-hello-part4/modules/local/cowpy/main.nf
vdauwera: Fixed

vdauwera:
@pinin4fjords I addressed your comments, lmkwyt about the diagrams.

(Also did a bunch of formatting updates and syntax updates while I was at it)

pinin4fjords:
> Hmmm I see your point. Would it be better if we showed one row highlighted?
> 
Yep, that'll do it!

PR #763, opened 2026-01-09
Update SQ and NF4Sci courses to strict syntax

vdauwera:
- [x] Side Quests
- [x] NF4 Science

Hello/Run courses covered in a separate branch


vdauwera:
The NF4Science courses still need some updates but the syntax update is done along with lots of formatting updates to be consistent with the newer styling, use of admonitions etc

PR #775, opened 2026-01-16
Improve motivation for workflow-level output{} block

pinin4fjords:
## Summary

Addresses feedback that the `output {}` block's verbosity compared to `publishDir` is not compellingly motivated in the training materials.

Uses framing from live teaching: "it's really the workflow's job to decide where outputs should be published. It shouldn't really be done by a process."

## Changes

| File | Section | Change |
|------|---------|--------|
| `docs/hello_nextflow/01_hello_world.md` | 2.4 (FYI) Process-level publishDir | Reframed benefits around separation of concerns and scattered logic problem |
| `docs/hello_nextflow/03_hello_workflow.md` | 1.5.2 Update the output block | Simplified trade-off explanation using "workflow's job" framing |

## Before

Part 1:
> "You will still find this code pattern all over the place in older Nextflow pipelines and process modules, so it's important to be aware of it. However, we do not recommend using it in any new work as it will eventually be disallowed in future versions of the Nextflow language."

Part 3:
> "Of course, that does mean we're repeating some information here, which could become inconvenient if we wanted to update the location for all the outputs in the same way."

## After

Part 1 now explains using "workflow's job" framing:
> **Why workflow-level outputs are better:**
> - **Separation of concerns**: Processes handle computation; workflows decide where outputs go. A process shouldn't dictate where its results are published—that's the workflow's job.
> - **Visibility**: With `publishDir`, output logic gets scattered across many files. In a complex pipeline, understanding where results end up requires searching through every process definition and config file. With workflow-level outputs, all publishing decisions are in one place.

Part 3 now uses concise trade-off explanation:
> "This does mean writing more lines than a process-level publishDir would require. The trade-off is worth it: deciding where outputs go is the workflow's job, not the process's, and having all output settings in one place makes the pipeline's structure easier to understand."

## Testing

- [x] No functional changes to code
- [x] Documentation reads clearly
- [x] Framing tested in live teaching session

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/docs/hello_nextflow/03_hello_workflow.md
+++ b/docs/hello_nextflow/03_hello_workflow.md
@@ -272,8 +272,9 @@ Once again, the logic is the same as before.
 This shows you that you can control the output settings at a very granular level, for every individual output.
 Feel free to try changing the paths or the publish mode for one of the processes to see what happens.
 
-Of course, that does mean we're repeating some information here, which could become inconvenient if we wanted to update the location for all the outputs in the same way.
-Later in the course, you'll learn how to configure these settings for multiple outputs in a structured way.
+This does mean writing more lines than a process-level `publishDir` would require.
+The trade-off is worth it: deciding where outputs go is the workflow's job, not the process's, and having all output settings in one place makes the pipeline's structure easier to understand.
+Later in the course, you'll see how to configure these settings for multiple outputs in a structured way.
 
 ### 1.6. Run the workflow with `-resume`
 
--- a/docs/hello_nextflow/01_hello_world.md
+++ b/docs/hello_nextflow/01_hello_world.md
@@ -588,7 +588,7 @@ We'll show you how to use them in due time in your Nextflow journey.
 
 ### 2.4. (FYI) Process-level `publishDir` directive
 
-Until very recently, the established way to publish outputs was to do it at the level of each individual process using a `publishDir` directive.
+Until recently, the established way to publish outputs was to do it at the level of each individual process using a `publishDir` directive.
 
 To achieve what we just did for the outputs of the `sayHello` process, we would have instead added the following line to the process definition:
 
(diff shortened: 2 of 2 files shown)


docs/hello_nextflow/03_hello_workflow.md
```diff
@@ -272,8 +272,9 @@ Once again, the logic is the same as before.
 This shows you that you can control the output settings at a very granular level, for every individual output.
 Feel free to try changing the paths or the publish mode for one of the processes to see what happens.
 
-Of course, that does mean we're repeating some information here, which could become inconvenient if we wanted to update the location for all the outputs in the same way.
-Later in the course, you'll learn how to configure these settings for multiple outputs in a structured way.
+This does mean writing more lines than a process-level `publishDir` would require.
```
bentsherman, line 275: At the same time, it's not really fair to compare the `output` block to a single `publishDir` directive. You really should compare it to the combination of all publishDir settings in your pipeline (like [this mostrosity](https://github.com/nf-core/rnaseq/blob/3.14.0/conf/modules.config))

I don't actually know which one is shorter. It likely depends on how you write the output declarations

Anyway, it might be better to frame it this way, as long as you don't come off as "um ackshually..." :nerd_face:

vdauwera:
Tbqh I'm not sure this argumentation belongs here. It's not really the beginner training's job to justify why we're moving to a newer syntax; that belongs in a migration guide, imho. I would prefer for the beginner training to focus on showing what people should do. If this explanation is already stated somewhere (as I assume it must be) perhaps we could just refer interested users to that?

bentsherman:
The Nextflow docs enumerate the benefits of workflow outputs here: https://nextflow.io/docs/latest/tutorials/workflow-outputs.html#overview

And code length is not one of them. Not that workflow outputs are necessarily longer or shorter, just that it's not really a top factor. It's as long as it needs to be

So you might be right that it's not even worth trying to argue about

ewels:
Agree that we don't need to justify ourselves. Or at least, _"`publishDir` will soon be deprecated and removed"_ should be enough justification.

pinin4fjords:
Fair enough, closing!

PR #776, opened 2026-01-19
Custom template style for course index pages

ewels:
Makes the markdown easier to write and pages easier to standardise.


vdauwera:
Oooh I like this a lot. Haven't had a chance to test yet (am afk) but overall it looks good.

PR #821, opened 2026-02-03
Add retry logic and parallelize translation workflow

ewels:
## Summary

Fixes the translation workflow failure from https://github.com/nextflow-io/training/actions/runs/21632620151 where an `APIConnectionError` caused the job to fail after ~2 hours.

**Changes:**
- Add exponential backoff retry logic to `call_claude()` - 8 retries with delays: 2s, 4s, 8s, 16s, 32s, 64s, 128s, 256s (~8.5 min total retry time)
- Set 10-minute request timeout per API call
- Parallelize translations using GitHub Actions matrix strategy - each language runs in its own parallel job
- Add `fail-fast: false` so other languages continue if one fails
- Add merge job to combine artifacts from all successful translations into a single PR
- Support partial success: PR is created even if some languages fail

**Benefits:**
- **Resilience**: Retry logic handles transient network/API failures
- **Speed**: 9 languages run in parallel instead of sequentially (~9x faster)
- **Partial success**: If 7/9 languages succeed, PR is created for those 7
- **No artifact clashes**: Each language has its own directory, no overlap possible

### DIFF
--- a/.github/workflows/translate.yml
+++ b/.github/workflows/translate.yml
@@ -50,28 +50,83 @@ jobs:
     needs: detect
     if: needs.detect.outputs.has_work == 'true'
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false # Continue other languages if one fails
+      matrix:
+        language: ${{ fromJson(needs.detect.outputs.languages) }}
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


ewels:
Cowboy merge coming up as I need to test this on master 🤠

PR #822, opened 2026-02-03
Quieten uv output and force colors in GitHub Actions

ewels:
## Summary
- Add `-q` flag to all `uv run` commands in GitHub Actions workflows to suppress package installation output
- Add `force_terminal` to Rich Console in `translate.py` to preserve colored output when running in CI


ewels:
Smol tweaks, 🤠 merge to self-test on master..

PR #835, opened 2026-02-04
Move dev version to bottom of version picker dropdown

ewels:
## Summary

- Moves the `0.dev` version to the bottom of the MkDocs Material version picker dropdown instead of the top
- New releases continue to appear at the top (most recent first)

The change will take effect on the next dev deployment, which will automatically reposition `0.dev` in `versions.json`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -181,17 +181,11 @@ jobs:
           }
 
           if is_release:
-              # Insert after any dev versions
-              insert_idx = 0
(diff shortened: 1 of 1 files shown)


ewels:
Logic was lost when the workflows were merged in https://github.com/nextflow-io/training/pull/819

All sites (dev and deployed releases) should update as soon as this is merged, as there will be a new dev deployment which will run this, and update the shared JSON file with the list of versions.

PR #838, opened 2026-02-05
Hello nextflow edits

ewels:
Mostly-minor edits to the "Hello Nextflow" training material that I spotted as I went through recording the new videos.

I doubt anyone except me will notice most of these, but still.


vdauwera (approved):
Hah your pedantry is appreciated. These all make sense to me. Couple of good catches in there too.

Issue #837, opened 2026-02-05
'Hello Nextflow' Suggestion: Remove shebangs


ewels:
In Hello Nextflow, we move processes into their own module files. We make special note to also copy over the shebang line (`#!/usr/bin/env nextflow`) to the module file, like so:

```nf
#!/usr/bin/env nextflow

process sayHello {

    input:
    val greeting

    output:
    path "${greeting}-output.txt"

    script:
    """
    echo '${greeting}' > '${greeting}-output.txt'
    """
}
```

In practice, this shebang makes basically zero difference though. We never execute these files from bash directly without the `nextflow` command interpreter. In the past it helped having it there to get syntax highlighting in some IDEs, but I think basically all recognise the `.nf` file extension now anyway.

I wonder if we can slightly simplify the material by just not mentioning the shebang and ditching it from modules.

Put this into an issue instead of a PR as I wasn't sure how controversial it'd be.
'cc @bentsherman @robsyme and others to keep me honest in case there's some technical reason to keep them that I haven't thought of.

vdauwera:
Hah I recall not having them originally and a pass being made of adding them systematically everywhere. No strong feelings on my part either way.

bentsherman:
The shebang is only useful for scripts that have an entry workflow, otherwise there is no need for it

PR #841, opened 2026-02-05
Remove module shebangs

ewels:
Script shebangs are not needed in modules, they're never directly executed. We only need them in Nextflow scripts that have an entry workflow (and to be honest, they're not _really_ needed there either).

Removed from all modules and removed from the "Hello Nextflow" training material where modules are created.

Fixes https://github.com/nextflow-io/training/issues/837

### DIFF
--- a/.claude/skills/run-tutorial/SKILL.md
+++ b/.claude/skills/run-tutorial/SKILL.md
@@ -9,6 +9,55 @@ Walk through a training tutorial lesson as a learner would, progressively buildi
 
 ---
 
+## Operating Modes
+
+This skill supports two distinct modes. **Ask the user which mode** if not clear from context:
+
+### Mode A: Learner Simulation (Progressive Build)
+
+Walk through as a learner would, editing starter files progressively. Use when:
+
+- Testing that tutorial instructions are complete and correct
+- Verifying the learning path works step-by-step
+
+### Mode B: Documentation Verification
+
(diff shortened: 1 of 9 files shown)


.claude/README.md
```diff
@@ -165,7 +165,8 @@ Claude will automatically use these skills when appropriate, or you can mention
 
 - DSL2 syntax only
 - UPPERCASE process names
-- Always include shebang: `#!/usr/bin/env nextflow`
+- Always include shebang: `#!/usr/bin/env nextflow` in scripts with an entry workflow
+- Do not include shebang in module files, i.e. `**/modules/*.nf`
```
maxulysse, line 169: subworkflows?

ewels: If I'm being pedantic, subworkflows _are_ modules.. And actually the directory structure is the same right? So I don't think that we need to confuse Claude with this.

ewels:
> I like shebangs as a cosmetic 'this is nextflow' sort of thing, but not enough to fight you on it.

That's actually how I explained their need when I recorded this last night - cosmetic, helping IDEs know which language syntax highlighting to use etc. But I didn't really buy the explanation myself 😅 Also, we don't use them in nf-core modules.

Mostly I found it was a bit of friction to repeatedly have to copy two things into a file instead of one, when the second part didn't make any practical difference.

Issue #839, opened 2026-02-05
Suggestion: More use of `-output-dir`


ewels:
One part of the updated training with Workflow Outputs that didn't _quite_ sit with me was the use of `outputDir` with `params.batch` for path customisation. I think / hope that the primary way folks will publish files will be with this CLI flag (shortened to `-o`). I generally feel that dynamic paths like the ones used for the examples belong in the `output` block instead of `outputDir` config. It mixes different functionality for a single CLI flag and I found it a bit confusing.

Relevant parts of the docs:

https://github.com/nextflow-io/training/blob/724787f45edcbd6f3c79ff9e0f4e283cd8b88ae1/docs/en/docs/hello_nextflow/06_hello_config.md?plain=1#L531

https://github.com/nextflow-io/training/blob/724787f45edcbd6f3c79ff9e0f4e283cd8b88ae1/docs/en/docs/hello_nextflow/06_hello_config.md?plain=1#L593

Personally I would rework this, show that the above pattern is possible (and concise, I agree), but put the `-o` flag as the primary way of setting output directories.

We could also use this as a way to show that more complex manipulations are possible here:

https://github.com/nextflow-io/training/blob/724787f45edcbd6f3c79ff9e0f4e283cd8b88ae1/docs/en/docs/hello_nextflow/06_hello_config.md?plain=1#L652

For example:

```nf
path { params.batch + "/" + sayHello.name }
```
or
```nf
path { "${params.batch}/${sayHello.name}" }
```

Wanted to see what others think first, as it's a relatively large change that would alter the flow of this part of the docs a bit.

'cc @vdauwera @bentsherman and others.

vdauwera:
I don't disagree... I settled on using the batch param opportunistically and as a bit of a last minute thing, with just the goal of showing a few aspects of how this stuff can be configured (not aiming to recommend a specific practice necessarily), and nobody pushed back during reviews so it stuck. 

So I'm not opposed to changing it, but I am however not keen to make any big changes at this juncture. 

That being said we're going to be more or less stuck with whatever we do here for the year, so if it's hugely objectionable, then let's change it, but quickly.

bentsherman:
If `params.batch` is meant to be the "base" output directory, then I would say to just set it to `outputDir` in the config:

```groovy
outputDir = params.batch
```

Or discard it entirely and just tell people to set `-output-dir` at runtime.

[...]

PR #842, opened 2026-02-05
Hello Config: more use of `-output-dir` (option B)

ewels:
Possible implementation of what I was suggesting in https://github.com/nextflow-io/training/issues/839

> [!WARNING]
> I haven't actually run this updated version yet, so the terminal outputs etc. are just written by hand.
> If we like this content I will do that before merge.

This PR is what I would say is the "complete" version, rewritten to follow the kind of flow that I'd expect and showing how to combine use of both `-o` and dynamic paths.

It is longer and a little more complex, so I'm putting it here as a draft to compare against a more lightweight change which will be in PR https://github.com/nextflow-io/training/pull/843

PR #843, opened 2026-02-05
Hello Config: more use of `-output-dir` (option A)

ewels:
Possible implementation of what I was suggesting in https://github.com/nextflow-io/training/issues/839

> [!WARNING]
> I haven't actually run this updated version yet, so the terminal outputs etc. are just written by hand.
> If we like this content I will do that before merge.

This PR is aiming to have a more minimal diff and edit than https://github.com/nextflow-io/training/pull/843 whilst still shifting the focus a little to `-output-dir`. It doesn't stack the concepts on top of one another quite so explicitly, but hopefully it still gets the point across.

### DIFF
--- a/docs/en/docs/hello_nextflow/06_hello_config.md
+++ b/docs/en/docs/hello_nextflow/06_hello_config.md
@@ -547,10 +547,10 @@ Make the following code changes in the workflow file:
     }
     ```
 
-Now run the pipeline with `-output-dir` to specify the output location:
+Now run the pipeline with `-output-dir`, or `-o` for short, to specify the output location:
 
 ```bash
-nextflow run hello-config.nf -output-dir results/outdir
+nextflow run hello-config.nf -output-dir new_results
 ```
 
 ??? success "Command output"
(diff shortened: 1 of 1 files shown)


docs/en/docs/hello_nextflow/06_hello_config.md
```diff
@@ -612,33 +567,84 @@ nextflow run hello-config.nf --batch outdir
...
+    ```
+
+This sets the output directory to `config_results/` plus the value of the `batch` parameter as a subdirectory.
+Now you can change the output location by setting the `--batch` parameter:
+
+```bash
+nextflow run hello-config.nf --batch my_run
+```
+
+This publishes outputs to `config_results/my_run/`.
```
maxulysse, line 634: I like how simple this is

ewels: This particular part is essentially unchanged from the current training material 😅 

https://training.nextflow.io/latest/hello_nextflow/06_hello_config/#213-run-the-pipeline

ewels:
@pinin4fjords prefers https://github.com/nextflow-io/training/pull/842 and I think I do too, so closing this. Can reopen if anyone disagrees.

vdauwera:
I agree option B is better

PR #847, opened 2026-02-06
Add examples of good writing in Polish

itrujnara:
Added examples of good writing in Polish to enhance the documentation.

PR #849, opened 2026-02-06
Detect prompt file changes in translation workflow

ewels:
## Summary

- Fixes translation workflow to re-translate all files when prompt files change
- Adds `prompt_changed_since()` function to detect changes to language-specific or general prompts
- Updates `ci-detect` and `sync` commands to show when prompt changes trigger full re-translation

## Problem

When PR #847 modified `docs/pl/llm-prompt.md`, the translation workflow ran but found no work to do. This happened because the workflow only checked if English source files changed since the baseline, not prompt files.

## Solution

Now when a language-specific prompt (`docs/{lang}/llm-prompt.md`) or the general prompt (`_scripts/general-llm-prompt.md`) changes, all existing translations for that language are marked as outdated and will be re-translated with the updated prompts.

Closes the gap identified after https://github.com/nextflow-io/training/pull/847

### DIFF
--- a/_scripts/translate.py
+++ b/_scripts/translate.py
@@ -352,6 +352,23 @@ def get_translation_baseline(
     )
 
 
+def prompt_changed_since(lang: str, baseline: str) -> bool:
+    """Check if any prompt files affecting this language changed since baseline.
+
+    Returns True if either:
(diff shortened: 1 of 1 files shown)


ewels:
Continuing my 🤠 behaviour with the language stuff. I feel we're getting closer though..

PR #867, opened 2026-02-09
Hello Nextflow: add warning for first time a v1 syntax parser error can happen

ewels:
Found the first `nextflow run` command in the training material that can throw an error with the old language parser and added an admonition to help people who might run into this.

Collapsable, looks like this when expanded:

CleanShot 2026-02-09 at 16 59 38@2x


ewels:
Updated to have a different admonition styling:

CleanShot 2026-02-09 at 18 07 25@2x

ewels:
GitHub issues are causing the Netlify builds to fail 🔥 

Should be fine to merge + deploy I think, on the code side.

ewels:
Added an equivalent warning to the "Nextflow Run" course:

CleanShot 2026-02-10 at 12 22 56@2x

PR #884, opened 2026-02-12
Workflow outputs in Side Quests

vdauwera:
## Summary                                                                                                                         
                                                                                                                                     
  This PR extends the Nextflow training side quests to teach the `publish:` / `output {}` workflow output mechanism introduced in Nextflow 24.x, replaces legacy `publishDir` usage, and significantly reworks several existing side quests.                         
                                                                                                                                     
  ---                                                                                                                                
                                                                                                                                     
  ## Major updates to Side Quests

  ### Workflows of Workflows (`side_quests/workflows_of_workflows/`)

  Significantly reworked to teach the `publish:` / `output {}` output mechanism and improve the overall lesson structure.
                                                                                                                                     
  - The lesson is restructured into two parallel sections (greeting workflow, then transform workflow), each following the same progression: review and run the standalone workflow → make it composable (`take:` / `emit:`) → wire it into `main.nf` using `include`, `publish:`, and `output {}`.                
  - Starter files added: `workflows/greeting.nf` and `workflows/transform.nf` are now pre-written standalone workflows that learners run first, then convert to composable named workflows.                                                                             
  - `main.nf` starter updated: pre-populated with `publish:` / `output {}` scaffolding (empty placeholder), progressively filled in during the lesson.                                                                                                                 
  - Solution `main.nf` updated: replaced `.view()` calls with a proper `publish:` / `output {}` block publishing all three outputs (`greetings`, `upper`, `reversed`) to named subdirectories; `nextflow.config` added.                                               
  - Bug fix: standalone `transform.nf` input path corrected from `results/timestamped/*.txt` to `results/timestamped_*.txt` to match actual filenames produced by the `TIMESTAMP_GREETING` process.                                                                     
                                                         
  ### Metadata (`side_quests/metadata/`)                                                                                             
                                                         
  Major rework to teach `publish:` / `output {}` as the output mechanism throughout, rather than introducing it only at the end.
                                                                                                                                     
  - The starter `main.nf` is now pre-populated with `publish:` / `output {}` scaffolding from section 1 onward, so learners use the mechanism from the first step.                                                                                                     
  - The per-step solution checkpoints have been fully regenerated and reorganized: `1.2`, `1.3`, `1.4`, `1.5`, `2.1`, `2.2`, `2.3`, `2.4`, `3.2` — each with consistent `main.nf` and module files reflecting the cumulative state at that point.                      
  - Section 2.4 (wire `ch_languages` into COWPY and update the output block) was split into two sub-steps for better pacing.
                                                                                                                                     
  ### Working with Files (`side_quests/working_with_files/`)

  Section 6 (processing files with metadata) migrated from `publishDir` to `publish:` / `output {}`:
                                                                                                                                     
  - `publishDir` directive removed from `ANALYZE_READS` process; output tuple changed from `val(meta.id)` to `val(meta)` so full metadata is available downstream.                                                                                                  
  - A dynamic `path` closure in the `output {}` block now controls the output directory structure: `path { meta, file -> 
  "${meta.type}/${meta.id}/${meta.replicate}" }`.                                                                                    
  - All doc snippets, Before/After blocks, and the summary section updated to reflect the new approach.
  - Minor doc improvements throughout: clarifying sentences, corrected console output (`class java.lang.String`, `class sun.nio.fs.UnixPath`), shebang removed from module snippet, a stray `}` bracket fixed.                                             
                                                                                                                                     
  ---                                                                                                                                
                                                         
  ## Navigation

  Side quests in `mkdocs.yml` reordered alphabetically within the section, with Plugin Development grouped as its own sub-section and Workflows of Workflows placed at the end of the list.

  ---

  ## Bug fixes and minor improvements (found during tutorial walkthroughs)
                                                                                                                                     
  - **Debugging**: `buggy_workflow.nf` starter and solution updated; `bad_resources.nf` and `nextflow.config` corrections.
  - **Essential Scripting Patterns**: solution `main.nf` and `nextflow.config` updated; `generate_report.nf` shebang removed from module.                                                                                                                            
  - **nf_test**: `main.nf`, `nextflow.config`, and test file corrections.
  - **Plugin Development**: `GreetingObserverTest.groovy` fixes in two solution steps.                                               
  - **Working with Files starter** (`working_with_files/main.nf`, `analyze_reads.nf`): aligned with the updated output mechanism.    
  - **IDE features** starter files (`basic_workflow.nf`, `complex_workflow.nf`, modules): minor corrections.


docs/en/docs/side_quests/debugging.md
```diff
@@ -2513,6 +2510,26 @@ Now it's time to put the systematic debugging approach into practice. The workfl
...
+
+            publish:
+            processed = processFiles.out
+            heavy = heavyProcess.out
+            files = handleFiles.out
+        }
+
+        output {
+            directory params.output
+            mode 'copy'
```
bentsherman: This syntax is from an old preview version. These settings should go in the config:

```groovy
outputDir = params.output
workflow.output.mode = 'copy'
```

docs/en/docs/side_quests/nf-test.md
```diff
@@ -162,6 +158,15 @@ You can see the full workflow code below.
...
         convertToUpper(sayHello.out)
+
+        publish:
+        greetings = sayHello.out
+        upper_greetings = convertToUpper.out
+    }
+
+    output {
+        directory 'results'
+        mode 'copy'
```
bentsherman: Same as above

bentsherman: Also, the outputs here need to be declared even if they don't require any custom settings:
```groovy
output {
    greetings {
        // path '.'
    }

    upper_greetings {
        // path '.'
    }
}
```

I commented the `path '.'` because that's what it will do by default, but you could add it if you want to be explicit / not have empty output declarations

docs/en/docs/side_quests/working_with_files/index.md
```diff
@@ -1596,9 +1592,8 @@ process ANALYZE_READS {
 
 !!! note
 
-    The `tag` and `publishDir` directives use closure syntax (`{ ... }`) instead of string interpolation (`"${...}"`).
-    This is because these directives reference input variables (`meta`) that aren't available until runtime.
-    The closure syntax defers evaluation until the process actually runs.
+    The `tag` directive uses closure syntax (`{ ... }`) because it references input variables (`meta`) that aren't available until the process runs.
+    The closure defers evaluation until runtime.
```
bentsherman, line 1612: Slightly off-topic -- with the v2 parser you don't need the closure here anymore, it's optional. If this guide is assuming the v2 parser, then it should be safe to remove the closure

But you might still want to have a note about the fact that the closure used to be needed

docs/en/docs/side_quests/working_with_files.md
```diff
@@ -1689,20 +1684,75 @@ Now let's actually call the `ANALYZE_READS` process on the `ch_samples` channel.
...
+           [
+               [
+                   id: sample,
+                   replicate: replicate.replace('rep', ''),
+                   type: type
+               ],
+               files
+           ]
+        }
+            .set { ch_samples }
```
bentsherman: Slightly off-topic -- try to prefer assignments over `set` and `tuple()` over `[]` (when making a tuple):

```groovy
        ch_samples = ch_files.map { id,  files ->
           def (sample, replicate, type, readNum) = id.tokenize('_')
           tuple(
               [
                   id: sample,
                   replicate: replicate.replace('rep', ''),
                   type: type
               ],
               files
           )
        }
```

This code will work better with the type checker in the future

docs/en/docs/side_quests/working_with_files.md
```diff
@@ -1689,20 +1684,75 @@ Now let's actually call the `ANALYZE_READS` process on the `ch_samples` channel.
...
         // Run the analysis
         ANALYZE_READS(ch_samples)
+
+        publish:
+        analysis_results = ANALYZE_READS.out
+    }
+
+    output {
+        directory 'results'
+        mode 'copy'
```
bentsherman: Same as above, move to config

docs/en/docs/side_quests/working_with_files.md
```diff
@@ -2021,7 +2075,14 @@ Applying these techniques in your own work will enable you to build more efficie
     - Organize outputs based on metadata
 
     ```groovy
-    publishDir { "results/${meta.type}/${meta.id}/${meta.replicate}" }, mode: 'copy'
+    output {
+        directory 'results'
+        mode 'copy'
```
bentsherman: Same as above, move to config

side-quests/debugging/buggy_workflow.nf
```diff
@@ -89,4 +86,24 @@ workflow {
...
+
+    publish:
+    processed = processFiles.out
+    heavy = heavyProcess.out
+    files = handleFiles.out
+}
+
+output {
+    directory params.output
+    mode 'copy'
```
bentsherman: Same as above, move to config

side-quests/debugging/buggy_workflow.nf
```diff
@@ -89,4 +86,24 @@ workflow {
 
     file_ch = channel.fromPath("*.txt")
     handleFiles(file_ch)
+
+    publish:
+    processed = processFiles.out
+    heavy = heavyProcess.out
+    files = handleFiles.out
```
bentsherman, line 93: Slightly off-topic -- I think we should try to use regular assignments instead of `.out` as much as possible, since `.out` is a relatively unusual syntax pattern whereas assignments are a widely understood concept

I think `.out` is mainly useful for when a process / workflow has multiple outputs, but in this case, each process has only one output and you're already assigning the first two (`processed_ch` and `heavy_ch`), so I think it's reasonable to avoid `.out` entirely here.

side-quests/ide_features/basic_workflow.nf
```diff
@@ -47,4 +47,20 @@ workflow {
     // View outputs
     FASTQC.out.html.view { "FastQC report: ${it}" }
     FASTQC.out.zip.view { "FastQC data: ${it}" }
+
+    publish:
+    fastqc_html = FASTQC.out.html
+    fastqc_zip = FASTQC.out.zip
```
bentsherman, line 53: (continuing discussion of `.out` vs assignments)

Here I think it's fair to use `.out`. Hopefully with records, this will collapse to a single channel of records, at which point we could remove the use of `.out` here as well.

side-quests/ide_features/basic_workflow.nf
```diff
@@ -47,4 +47,20 @@ workflow {
...
+}
+
+output {
+    directory params.output_dir
+    mode 'copy'
```
bentsherman: Same as above, move to config

side-quests/ide_features/basic_workflow.nf
bentsherman: Actually, what I would go ahead and do here is to join the two channels into a single channel:

```groovy
    ch_fastqc = FASTQC.out.html.join(FASTQC.out.zip)

    publish:
    fastqc = ch_fastqc
```

Then you only have one output declaration:
```groovy
output {
    fastqc {
        path 'fastqc'
    }
}
```

My intuition is that it will be better to combine related outputs into a single workflow output as much as possible

You could imagine taking this further and combining all "per-sample" outputs into a single workflow output called `samples`, while using the dynamic path closure to keep organizing output files by tool. That is more a matter of personal taste, but at the very least I think we should at least collapse to one workflow output per process output, as a best practice.

bentsherman: You could also convert the tuples to maps, in anticipation of records:
```groovy
    ch_fastqc = FASTQC.out.html.join(FASTQC.out.zip)

    publish:
    fastqc = ch_fastqc.map { sample_id, html, zip -> [sample_id: sample_id, html: html, zip: zip] }
```

But I wouldn't say this is necessary at this point

side-quests/ide_features/complex_workflow.nf
```diff
@@ -85,4 +87,32 @@ workflow {
...
+    }
+    trimmed {
+        path 'trimmed'
+    }
+    alignments {
+        path 'alignments'
+    }
+    counts {
+        path 'counts'
+    }
```
bentsherman, line 111: (continuing discussion of collapsing outputs)

Just to illustrate, this is what it would like to collapse outputs to the extreme:
```groovy
    samples { s ->
        s.fastqc_html >> 'fastqc/'
        s.fastqc_zip >> 'fastqc/'
        s.trimmed_fastq >> 'trimmed/'
        s.alignments >> 'alignments/'
        s.counts >> 'counts/'
    }
```

This assumes you join the individual output channels into a single channel of maps / records:
```groovy
    samples = RNASEQ_PIPELINE.out.fastqc
        .join( RNASEQ_PIPELINE.out.trimmed )
        .join( RNASEQ_PIPELINE.out.alignments )
        .join( RNASEQ_PIPELINE.out.counts )
        .map { sample_id, fastqc_html, fastqc_zip, trimmed_fastq, alignments, counts -> [sample_id: sample_id, fastqc_html: fastqc_html, fastqc_zip: fastqc_zip, trimmed_fastq: trimmed_fastq, alignments: alignments, counts: counts] }
```

As you can see, it gets pretty ugly when you're still using tuples... so I'm not suggesting that we do this now. Just something to consider for when we eventually introduce records

bentsherman:
I think that concludes my review. Main two points are:

1. `directory` and `mode` in the `output` block need to be moved to the config
2. every output assigned in the `publish:` section needs to be declared in the `output` block

Aside from that, I left some extra thoughts about best practices. Feel free to take them or leave them

Hopefully you can just throw my general suggestions into a spec and let the AI update all of the code examples for you :smile:

vdauwera:
> I think that concludes my review. Main two points are:
> 
> 1. `directory` and `mode` in the `output` block need to be moved to the config
> 2. every output assigned in the `publish:` section needs to be declared in the `output` block
> 
> Aside from that, I left some extra thoughts about best practices. Feel free to take them or leave them
> 
> Hopefully you can just throw my general suggestions into a spec and let the AI update all of the code examples for you 😄

Thanks @bentsherman, this is super helpful!

side-quests/solutions/essential_scripting_patterns/nextflow.config
```diff
@@ -1,3 +1,6 @@
 // Nextflow configuration for Groovy Essentials side quest
 
 docker.enabled = true
+
+outputDir = 'results'
```
bentsherman, line 5: For what it's worth, the output dir defaults to `results`, so you could omit this setting. But I think it's totally fair if you want to be explicit about it

side-quests/debugging/buggy_workflow.nf
pinin4fjords: Addressed in the solution file (side-quests/solutions/debugging/buggy_workflow.nf), which now uses variable assignments (processed_ch, heavy_ch, file_ch) in the publish block instead of .out. The buggy starter file intentionally keeps .out since it represents the pre-fix state.

side-quests/ide_features/basic_workflow.nf
pinin4fjords: Acknowledged. .out is kept here as suggested. The follow-up suggestions (joining channels, converting to maps) were noted as not necessary at this point.

side-quests/ide_features/complex_workflow.nf
pinin4fjords: Acknowledged as a future consideration for when records are introduced. No changes needed now as noted.

docs/en/docs/side_quests/working_with_files/index.md
pinin4fjords: This one is still open. The docs and module still use `tag { meta.id }` (closure syntax) and the docs include a note defending the closure. With the v2 parser, `tag meta.id` works without braces. @vdauwera - do you want to update this to drop the closure and adjust the note, or leave as-is for now?

vdauwera: Whatever you think is best, I don't have a strong opinion about it. Maybe leave as is and refine later, in the interest of getting this out?

pinin4fjords: OK!

mribeirodantas (approved):

PR #910, opened 2026-03-12
Update 01_basics.md

fraserclaire:
Changed "Pipeline parameters" in code block comments to "파이프라인 매개변수"


ewels:
Please don't update the docs themselves - these files will be overwritten the next time that the translation jobs run!

See https://github.com/nextflow-io/training/blob/master/TRANSLATING.md#why-not-edit-translations-directly

PR #913, opened 2026-03-12
Update 03_config.md

fraserclaire:
Replaced English code block comments with Korean translation


ewels:
Please don't update the docs themselves - these files will be overwritten the next time that the translation jobs run!

See https://github.com/nextflow-io/training/blob/master/TRANSLATING.md#why-not-edit-translations-directly

PR #915, opened 2026-03-12
Enhance Turkish translation guidelines

sudekaraoglan:
This PR improves the Turkish translation guidelines with a focus on sentence structure and workflow clarity. Key changes include:

- Added new rules for handling complex phrases and long introductory sentences in Turkish.

- Updated TRANSLATING.md diagrams and text to accurately reflect the current translation workflow.

- General refinements to the translation guidelines for better readability and consistency.

PR #917, opened 2026-03-13
Automated turkish translation updates

ewels:
Follows on from https://github.com/nextflow-io/training/pull/915


ewels:
@sudekaraoglan - please take a look. As before, if it's an improvement we should try to merge fairly soon. Can always do more follow-up PRs to the prompt if needed.

sudekaraoglan:
Hi @ewels - I've checked the changes. It's definitely an improvement and looks much better now. I'll continue to work on minor refinements in future PRs, but this is good to go for now. Thanks!

PR #914, opened 2026-03-12
Update Korean llm prompt

fraserclaire:
Update Korean LLM translation prompt with instructional style and terminology rules

PR #918, opened 2026-03-13
Updated docs/ko/llm-prompt.md

jhlee0637:
As the result of the nf-core Hacakthon 2026 March from Seoul local hub, I have merged all the PRs from the attendees. 

We have added some translation/transliteration.

I considered updating `glossary.yml` also, but to make it simple we updated the `llm-prompt.md` file only.

### DIFF
--- a/docs/ko/llm-prompt.md
+++ b/docs/ko/llm-prompt.md
@@ -118,67 +118,66 @@ Note: Some terms use transliteration (음차) while others use actual Korean tra
 
 ### Transliterations (음차)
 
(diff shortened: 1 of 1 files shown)


ewels:
Thanks for this @jhlee0637!

How do you want to handle this alongside https://github.com/nextflow-io/training/pull/914 ?

jhlee0637:
Hi @ewels  Thank you for merging them!
I think it's okay to merge it with #914 now, I'd like to see how the translation becomes changed later with it.

PR #761, opened 2026-01-09
Add plugin development side quest, restructure side quests for multi-part courses

pinin4fjords:
## Summary

Adds a **Plugin Development** side quest (6 progressive lessons on Nextflow plugin development) and restructures all side quests to support multi-part courses consistently.

### Plugin Development course

- **Part 1**: Using existing plugins (nf-hello, nf-schema, nf-co2footprint), including config demos
- **Part 2**: Scaffolding a plugin project with `nextflow plugin create`
- **Part 3**: Custom @Function methods (reverseGreeting, decorateGreeting)
- **Part 4**: Spock unit testing
- **Part 5**: TraceObserver lifecycle events (GreetingObserver, TaskCounterObserver)
- **Part 6**: Configuration (navigate() for reading, ConfigScope for declaring)

Written for Nextflow 25.10.2 using the plugin Gradle plugin (io.nextflow.nextflow-plugin 1.0.0-beta.10) and config.spec.* API.

### Side quests restructure

- Each side quest now lives in its own subdirectory (`side_quests/questname/index.md`), matching the nf4_science pattern and allowing any quest to expand into multiple pages in the future
- Single-page quests remain direct nav entries (no extra click); multi-part courses like Plugin Development get a named subsection
- Updated all cross-references between side quest pages for new paths
- Added Plugin Development course box to the side quests index page

### Other changes

- `.editorconfig`: Makefile tab rule for plugin solutions
- `.gitignore`: `.gradle/` ignore for plugin builds
- `.pre-commit-config.yaml`: Excludes for template-generated files (COPYING, gradlew)
- `.prettierignore`: Exclude COPYING files in plugin solutions
- `docs/en/docs/envsetup/02_local.md`: Java 21 requirement note for plugin development

## Test plan

- [x] Preview locally with MkDocs
- [x] Tutorial walkthrough tested for Parts 1-6 (Mode A, learner simulation in Docker)
- [x] All solutions build and run correctly
- [x] All solutions pass `-resume` caching
- [x] Validation checks pass on all 6 lesson files (headings, hl_lines, inline code, writing style)
- [x] Nav structure verified via deploy preview

## Follow-up issues

- #921 - Consider removing/moving mermaid class diagram in Part 2
- #922 - Tighten config section (Part 6)
- #923 - Add meaningful TraceObserver data usage example (Part 5)
- #925 - Coordinate with Nextflow docs to avoid content overlap

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #940, opened 2026-04-27
Fix broken image links in dev_environment and debugging side quests

robsyme:
## Summary

The restructure in #761 moved `ide_features.md` → `dev_environment/index.md` and `debugging.md` → `debugging/index.md`, but the embedded image references stayed at `img/...`. With the new directory depth, that path resolves to `side_quests/dev_environment/img/...` (which doesn't exist) instead of `side_quests/img/...` (where the images live), so every screenshot on those two pages 404'd on the published site — e.g. https://training.nextflow.io/0.dev/side_quests/dev_environment/img/outline.png.

Fix is a mechanical `](img/` → `](../img/` across the 28 affected references on the two pages, applied to English and all 10 translations.

- `dev_environment/index.md`: 24 references
- `debugging/index.md`: 4 references

## Test plan

- [x] Reproduced 404s locally with `uv run _scripts/docs.py live --lang en --dirty` against the broken state
- [x] Verified all 27 unique image URLs return 200 after the fix
- [x] Confirmed rendered HTML now uses `src="../img/..."` resolving to `/side_quests/img/...`
- [ ] CI build passes
- [x] Spot-check pages on the deploy preview


vdauwera (approved):
Thanks for the fix!

PR #944, opened 2026-04-28
Branch to accumulate minor fixes based on Summit trainings

vdauwera:
Should only be doc fixes, nothing functional (use a separate branch for functional changes)


vdauwera:
reused a previous branch name by accident

PR #945, opened 2026-04-28
Branch to accumulate minor fixes from Summit training

vdauwera:


vdauwera:
These are all trivial fixes made by claude which i've double-checked and feel confident in.

PR #949, opened 2026-05-08
Stop search engines indexing old versioned doc paths

ewels:
Generate a `_headers` file at the gh-pages root from `versions.json` during deploy, sending `X-Robots-Tag: noindex` for every version path except `/latest/*`.

This deindexes URLs like `https://training.nextflow.io/2.0/...`, `/0.dev/...`, `/development/...`, `/_dev/...` from Google while leaving the canonical `/latest/...` URLs indexable.

Also expanded the `netlify.toml` comment to make the deploy split explicit.

The `_headers` rule list is rebuilt from `versions.json` on every deploy, so it stays correct as new releases are published or aliased.

I just pushed this file to the `gh-pages` branch in https://github.com/nextflow-io/training/commit/4420c5647d2c727078e69764be8f5911acb6431e  - this PR adds the automation to rebuild the file every time there's a new release.


pinin4fjords (approved):
seems legit

PR #960, opened 2026-06-08
Fix gremlins in nf-test, essential scripting, and debugging side quests

pinin4fjords:
Walked the nf-test, Essential Scripting Patterns, and Debugging side quests end-to-end as a learner (latest master, Nextflow 25.10.4, strict v2 parser) and fixed the discrepancies found.

## Debugging
- Corrected workflow could time out: `heavyProcess` looped `{1..1000000}` while the only documented fix raised the limit to `100 s`. 1M iterations can exceed 100 s, so following the lesson verbatim left a failing workflow. Reduced to `{1..10000}` (matching the solution) in both the "Complete Corrected Workflow" block and the starter `buggy_workflow.nf`. The `time '1 ms'` bug is preserved for teaching; only the post-fix run is corrected.
- Indented the `exhausted.nf` command-output block so it renders inside its `??? success` collapsible.
- Aligned the no_such_var Before-tab output filename (`_processed.txt`), the "expects only 1 input channel" comments, and the `bad_resources.nf` snippet (added the omitted `container` line, fixed `hl_lines`) with the actual files.

## Essential Scripting Patterns
- Corrected the `.filter()` prose: with `depth >= 25000000` no samples are excluded (the lowest sample is exactly 25000000), so all three pass.
- Fixed the `generate_report.nf` compile-error line number to `13:27`.
- Fixed the `collect.nf` tab titles (were `main.nf`); used `tuple(...)` in the FASTP Before tab.
- Filled the missing run commands and the three empty command-output blocks (which rendered literal `<!-- TODO: output -->` text) with captured output; removed stray author-note placeholders.

## nf-test
- Regenerated the `convertToUpper` solution snapshot: its key did not match the test name, so the snapshot was silently regenerated instead of verified. Key and md5 now match.
- Added `nextflow.config` to the directory listing; aligned the "Workflow code" snippet with `main.nf` (shebang, flush input/output declarations).

Headings validate; no new prettier issues.

### Follow-up (not in this PR)
At Essential 2.3 the prose says reports land in `results/reports/`, but at that point the module has no `publishDir`/output block, so they go to `work/` until the output block is introduced later in the lesson. Flagging for a separate decision.

### DIFF
--- a/docs/en/docs/side_quests/debugging/index.md
+++ b/docs/en/docs/side_quests/debugging/index.md
@@ -512,7 +512,7 @@ If you get a 'No such variable' error, you can fix it by either defining the var
         val sample_name
 
         output:
-        path "${sample_name}_output.txt"
+        path "${sample_name}_processed.txt"
 
         script:
         // Define variables in Groovy code before the script
@@ -887,7 +887,7 @@ The error message clearly states that the call expected 1 argument but received 
 
 process PROCESS_FILES {
     input:
-        val sample_name  // Process expects only 1 input
+        val sample_name  // Process expects only 1 input channel
 
     output:
         path "${sample_name}_output.txt"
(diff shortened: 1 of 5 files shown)


FriederikeHanssen (approved):

pinin4fjords:
Thanks @FriederikeHanssen !

PR #963, opened 2026-06-15
[ DO NOT MERGE ] Side quest/co scientist

FriederikeHanssen:
PR currently purely to make commenting easier. Do not merge to master!

PR #968, opened 2026-06-16
Add Seqera Platform Automation side quest

adamrtalbot:
## Summary

Adds a new **Seqera Platform Automation** side quest that walks the Platform's programmatic surface across three workspace roles of decreasing privilege (Admin → Maintain → Launch), each handing an artifact to the next.

- **Lesson**: `docs/en/docs/side_quests/platform_automation/index.md` — create compute environments (Batch Forge UI + Terraform), add pipelines to the Launchpad four ways (UI, `tw`, seqerakit, Terraform), and launch (GUI, `tw`, seqerakit, Actions). The throughline is declarative existence (Terraform) vs imperative actions (UI, `tw`, seqerakit).
- **Runnable assets**: `side-quests/platform_automation/` — `terraform/compute-env/`, `terraform/pipeline/`, and `seqerakit/` with `*.tfvars.example` templates.
- **Devcontainer**: adds Terraform (`hashicorp.terraform` extension), `tw` (0.32.0), and `seqerakit` (0.5.7) to the toolchain.

## Notes

- The page is intentionally **not** wired into the nav / `side_quests/index.md` table in this PR.
- The "What's next?" link points at the **Building with Seqera AI (CoScientist)** side quest, which lands in #963. The relative link target only resolves once #963 merges — worth coordinating merge order.

## Verification

- `uv run .github/check_headings.py docs/en/docs/side_quests/platform_automation/index.md` → 0 errors
- `npx prettier --check` on the lesson → passes
- `git check-ignore` confirms `terraform.tfvars`/`*.tfstate` are ignored while `*.tfvars.example` stay tracked

Generated by Claude Code


docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+- Create a compute environment two ways, with increasing control over the cloud: the UI (Batch Forge) and Terraform
+- Add a pipeline to the Launchpad declaratively with Terraform, and see idempotency
+- Launch a pipeline using the GUI, CLI and `seqerakit`.
+- Tell declarative existence (Terraform) apart from imperative actions (the UI, `seqerakit`, `tw`) and learn when to use each
+
+### Prerequisites
+
+Before taking on this side quest, you should:
+
+- Have a Seqera Platform account on Seqera Cloud (`https://cloud.seqera.io`), or an Enterprise install
```
luisas:
```suggestion
- Have a Seqera Platform account on Seqera Cloud [https://cloud.seqera.io](`https://cloud.seqera.io`), or an Enterprise install
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+- Be a member of a workspace with a role: Admin, Maintain, or Launch
+- Have credentials for your cloud provider already added to that workspace (not covered here)
+- Ideally, a GitHub access token added too, to avoid GitHub rate limits
```
luisas:
```suggestion
- Ideally, GitHub credentials are added to your workspace. This is optional, but it would help prevent incidents caused by GitHub rate limits.
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+- Launch a pipeline using the GUI, CLI and `seqerakit`.
+- Tell declarative existence (Terraform) apart from imperative actions (the UI, `seqerakit`, `tw`) and learn when to use each
+
+### Prerequisites
+
+Before taking on this side quest, you should:
+
+- Have a Seqera Platform account on Seqera Cloud (`https://cloud.seqera.io`), or an Enterprise install
+- Be a member of a workspace with a role: Admin, Maintain, or Launch
+- Have credentials for your cloud provider already added to that workspace (not covered here)
```
luisas:
```suggestion
- Have credentials for your cloud provider already added to that workspace (not covered here, see [credentials overview](https://docs.seqera.io/platform-cloud/credentials/overview))
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+- Ideally, a GitHub access token added too, to avoid GitHub rate limits
+- Be comfortable with the command line and basic Nextflow concepts
+
+New to running pipelines on Seqera at all? Start with the gentler "Run pipelines on Seqera" module in the [Nextflow Triathlon](https://training.nextflow.io/) (sign up, launch in the UI, the `tw` CLI). This side quest is the automation layer on top of it: roles, Terraform, `seqerakit`, and Actions.
```
luisas: I cannot fin the triathlon in this link atm

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+    │       ├── main.tf
+    │       ├── variables.tf
+    │       └── terraform.tfvars.example
+    └── seqerakit              # sections 2 & 3: add and launch a pipeline
+        ├── add-rnaseq.yml
+        ├── launch-rnaseq.yml
+        └── launch-rnaseq-multiple.yml
+    ```
+
+### Create an access token
```
pinin4fjords, line 102:
```suggestion
### Log in to Platform and go to the organisation and workspaces your trainer directs you to

.... or something

### Create an access token
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+We authenticate to the Platform to tell it who we are. That is what grants the permissions of our role. Seqera uses a **personal access token**: a bearer token you create once and send with every API call, in an `Authorization: Bearer <token>` header. The token carries your identity and role, so the Platform knows who you are and what you can do. Anyone with the token can act as you, so keep it secret. In this training we save it as an environment variable.
+
+1. In the Platform, open the user menu (top right) and choose **Your tokens**.
+2. Click **Add token**, name it (e.g. `platform-automation`), and click **Add**.
+3. Copy the token now. The Platform shows it only once.
+4. In the Codespace terminal, export it under both names:
+
+```bash
+export TOWER_ACCESS_TOKEN=<paste-token>
+export SEQERA_ACCESS_TOKEN=$TOWER_ACCESS_TOKEN
```
pinin4fjords: Why both?

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+tw info
+```
+
+This prints the API endpoint and the authenticated user. If it errors, the token is wrong or not exported.
+
+!!! warning
+
+    This only exists in a single terminal session, if you open a new terminal you will need to export them again.
+
+For Seqera Enterprise, also set `SEQERA_API_URL` (and the `server_url` Terraform variable) to your install's API URL. Everything else is identical.
```
pinin4fjords: Example? How do folks find that?

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+| [Maintain](#2-maintain-add-a-pipeline-to-the-launchpad) | Maintain      | Pipelines on the Launchpad               | a Launchpad pipeline |
+| [Launch](#3-launch-run-a-pipeline)                      | Launch        | Nothing; can only run                    | pipeline runs        |
+
+Admin users create compute environments and assign roles to everyone else. If you only have Maintain, an Admin on your team hands you a `compute_env_id`; start at section 2. If you only have Launch, they hand you a Launchpad pipeline (and optionally an Action) to run; start at section 3. Each section opens with what it requires.
+
+---
+
+## 1. Admin: compute environments and cloud
+
+**Requires:** Owner or Admin role, and permissions in the cloud environment. If you are using a Cloud provider, authenticate with them first via their CLI.
```
pinin4fjords: We're skipping over credentials here. Don't we want to 1) ensure that trainers do that in advance, so that keys are in place and 2) have the users go look at the creds, understand what they are?

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+**AWS**
+
+- Region
+- Work directory (S3 bucket)
+- Wave, Fusion, Fast Instance should all be enabled
+- Config mode should be Batch Forge
+- Provisioning model should be Spot, using Spot instances for nodes which run Nextflow tasks.
+
+**Azure**
+
```
pinin4fjords, line 220: Lots of ways this can go wrong- should we encourage some sensible values?

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+- For the worker pool, select a VM type you have sufficient quota for. Standard_D4s_v3 (4 CPUs) is a good starting choice.
+- Set the worker pool VM count to 4
+- Leave autoscaling enabled for both pools
+
+Leave the other settings at their defaults. Keep **Dispose resources** enabled so the pool is torn down when you delete the compute environment. Click **Add** in the top right.
+
+Watch your cloud console and you will see Forge create the resources: an identity and roles for Nextflow (access to blob storage and the Batch service), a more limited role for the worker tasks (storage only), the pools, and their networking. Many moving parts, all handled for you.
+
+That convenience is the trade-off: Forge owns those resources and gives you little control. A lab that already runs on managed infrastructure wants the opposite, to describe the resources itself and wire the Platform to them. That is the Terraform path.
+
```
pinin4fjords, line 250: For our current example workspace:

```
Unable to connect - java.net.UnknownHostException: seqeralabs.westeurope.batch.azure.com: Name or service not known
```

docs/en/docs/side_quests/platform_automation/index.md
luisas:
```suggestion
**Requires:** Owner or Admin role, and permissions in the cloud environment. If you are using a Cloud provider, authenticate with them first via their CLI (e.g. `aws sso login`).
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+We authenticate to the Platform to tell it who we are. That is what grants the permissions of our role. Seqera uses a **personal access token**: a bearer token you create once and send with every API call, in an `Authorization: Bearer <token>` header. The token carries your identity and role, so the Platform knows who you are and what you can do. Anyone with the token can act as you, so keep it secret. In this training we save it as an environment variable.
+
+1. In the Platform, open the user menu (top right) and choose **Your tokens**.
+2. Click **Add token**, name it (e.g. `platform-automation`), and click **Add**.
+3. Copy the token now. The Platform shows it only once.
+4. In the Codespace terminal, export it under both names:
+
+```bash
+export TOWER_ACCESS_TOKEN=<paste-token>
```
pinin4fjords, line 112: we should probably have them store this somewhere persistent. I just lost mine because I had to resume my codespace and the env wasn't preserved

pinin4fjords:
```suggestion
echo <paste token> > token.txt
export TOWER_ACCESS_TOKEN=$(cat token.txt)
```

pinin4fjords: same for all the env stuff

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+        "disposeOnDeletion": true
+      }
+    }
+  }
+}
+```
+
+The `forge` block is exactly what the checkboxes set in 1.1. Its presence is what tells the Platform to create and scale the pools for you. In the next section, the absence of this block is what makes the Terraform compute environment manual.
+
+Close the loop: import the file to recreate the compute environment, with nothing retyped.
```
pinin4fjords:
```suggestion
This file can be used to recreate the compute environment, with nothing retyped.
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+    │   │   └── terraform.tfvars.example
+    │   └── pipeline           # section 2 (Maintain): add a pipeline to the Launchpad
+    │       ├── main.tf
+    │       ├── variables.tf
+    │       └── terraform.tfvars.example
+    └── seqerakit              # sections 2 & 3: add and launch a pipeline
+        ├── add-rnaseq.yml
+        ├── launch-rnaseq.yml
+        └── launch-rnaseq-multiple.yml
+    ```
```
pinin4fjords, line 94:
```suggestion
    ```

We'll cover the content of these files in a moment.
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+Terraform manages cloud resources declaratively: you describe what should exist and Terraform makes it so. It does not run your work; it only creates the compute environment. `side-quests/platform_automation/terraform/compute-env/` does it in a single `main.tf`, two providers in one apply:
+
+- data sources: the existing Batch account and the Azure credentials already
+  stored in the Platform, referenced, not created.
+- `azurerm`: creates a head pool and a worker pool on that account.
+- `seqera`: creates the compute environment pointing at those pools.
+
+The manual marker: `head_pool` is set to the pool Terraform just made and there is **no `forge` block**. That one difference is what makes the compute environment manual instead of Forge. `nextflow_config` routes tasks to the worker pool.
+
+Let's walk the key blocks of `terraform/compute-env/main.tf`.
```
pinin4fjords: Too much 'just talking' about tf below, I'd hate to teach that and just waffle on about terraform code blocks for 5 mins straight. We should have folks running pieces more, e.g. do the 'init' up front.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+```
+
+Terraform reads those references and works out the order itself: pools first, then the compute environment that depends on them.
+
+Now, we can run Terraform to apply these changes.
+
+```bash
+# you may need to log in to the cloud provider with `az login`
+terraform init
+terraform apply
```
pinin4fjords, line 419: This asks the user for a whole bunch of stuff we haven't told them how to find out.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+
+The compute environment now exists in two places, and as an Admin with cloud access you can see both.
+
+On the Platform side, read the ID Terraform produced:
+
+```bash
+terraform output compute_env_id
+```
+
+On the Azure side, open the Batch account in the portal. You will see the head and worker pools Terraform created, sitting idle with no jobs yet. Once you launch a pipeline (sections 2 and 3), jobs and tasks stack onto these pools, and you can drill into a task's logs and exit code. That is the whole point of the Admin tier: the Platform submits to Azure Batch, and Azure Batch runs the work. Maintain- and Launch-role users see only the compute environment in the workspace, not the cloud behind it.
```
pinin4fjords: Does this make sense to include? Are we ever going to give trainees access to cloud consoles?

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: Let's remove!
```suggestion
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+
+- **Name**: `rnaseq-nf-` followed by your handle, e.g. `rnaseq-nf-ada` (the handle you exported as `WORKSHOP_USER` in section 0).
+- **Compute environment**: the one from section 1, e.g. `azure-batch-manual`.
+- **Pipeline to launch**: `https://github.com/nextflow-io/rnaseq-nf`.
+- **Revision**: `master`.
+- **Work directory**: your Azure Blob work dir, e.g. `az://work`.
+
+Click **Add**. The pipeline appears on the Launchpad with no run started. Adding a pipeline only saves a launch configuration; it does not run anything.
+
+To remove from the workspace, you can click the hamburger menu on the top right and click **Delete**.
```
pinin4fjords: hamburger?

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,740 @@
...
+
+### 2.2. Add a pipeline via `tw`
+
+The CLI does the same thing in one command:
+
+```bash
+tw pipelines add \
+  --name="rnaseq-nf-$WORKSHOP_USER" \
+  --compute-env="azure-batch-manual" \
```
pinin4fjords, line 479: Will we guarantee 'azure-batch-manual'? What to do if we're not doing 'admin'?

pinin4fjords (commented):
Just submitting intermediate review while I continue

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+
+To remove from the workspace, you can click the hamburger menu on the top right and click **Delete**.
+
+### 2.2. Add a pipeline via `tw`
+
+The CLI does the same thing in one command:
+
+```bash
+tw pipelines add \
+  --name="rnaseq-nf-$WORKSHOP_USER" \
```
pinin4fjords, line 478:
```suggestion
  --name="rnaseq-nf-$WORKSHOP_USER-tw" \
```

prevent clash with manual creation

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+
+Run it again and `tw` errors: a pipeline with that name already exists. The command is imperative, so each invocation tries to add a pipeline; it has no notion of "already in the desired state".
+
+To remove the pipeline, you can use the `tw` command line again:
+
+```bash
+tw pipelines delete --name="rnaseq-nf-$WORKSHOP_USER"
+```
+
+### 2.3. Add a pipeline with `seqerakit`
```
pinin4fjords: Is it even worth teaching `seqerakit` with the Terraform provider available now?

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+
+`seqerakit` is a wrapper over `tw` that reads a YAML file and runs the underlying `tw` commands. It keeps the configuration as code, so a teammate can reproduce the exact same pipeline. `seqerakit/add-rnaseq.yml` describes the pipeline:
+
+```yaml
+pipelines:
+  - name: "rnaseq-nf-${WORKSHOP_USER}"
```
pinin4fjords:
```suggestion
  - name: "rnaseq-nf-${WORKSHOP_USER}-sk"
```

Prevent clashes

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+    url: "https://github.com/nextflow-io/rnaseq-nf"
+    workspace: "${TOWER_WORKSPACE_ID}"
+    description: "Added with seqerakit"
+    compute-env: "azure-batch-manual"
```
pinin4fjords:
```suggestion
    compute-env: "<CE name>"
```

unless guaranteed

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+```
+
+You can force it through with `seqerakit add-rnaseq.yml --overwrite`, which deletes and recreates the pipeline. But that is you telling it to repeat the action. By default, adding twice is an error. To make "add once, and leave it alone after that" the _default_ behaviour, we need a tool that manages existence rather than actions: Terraform.
+
+### 2.4. Add a pipeline with Terraform
+
+`side-quests/platform_automation/terraform/pipeline` adds the same pipeline declaratively. You describe the pipeline that should exist; Terraform makes the workspace match. Like the imperative methods above, the config tracks the `master` branch; you will pin it to a release tag below to see Terraform update the pipeline in place.
+
+```bash
+cd /workspaces/training/side-quests/platform_automation/terraform/pipeline
```
pinin4fjords, line 498: Not sure why, but I can't make this work

```
seqera_pipeline.rnaseq_nf: Creating...
╷
│ Error: unexpected response from API. Got an unexpected response code 403
│ 
│   with seqera_pipeline.rnaseq_nf,
│   on main.tf line 27, in resource "seqera_pipeline" "rnaseq_nf":
│   27: resource "seqera_pipeline" "rnaseq_nf" {
│ 
│ **Request**:
│ POST /pipelines?workspaceId=75541431575041 HTTP/1.1
│ Host: api.cloud.seqera.io
│ Accept: application/json
│ Authorization: (sensitive)
│ Content-Type: application/json
│ User-Agent: speakeasy-sdk/terraform 0.40.1 2.903.0 1.153.0
│ github.com/seqeralabs/terraform-provider-seqera/internal/sdk
│ 
│ 
│ **Response**:
│ HTTP/2.0 403 Forbidden
│ Content-Length: 0
│ Date: Thu, 18 Jun 2026 15:28:47 GMT
│ 
│ 
```

pinin4fjords (commented):
Just a few specific points below. General ones:

1. Should try and make sure everything is runnable without having to think too hard, make sure different flavours of the same commands don't conflict.
2. Is it worth teaching Seqerakit, or should we just do TF for infra and `tw` for e.g. pipeline runs? I've never found seqerakit's IaC to make much sense for runs. 
3. Need screenshots for Platform pieces.

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: it's a bug, not related to this. It still works fine.

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: yup: https://en.wikipedia.org/wiki/Hamburger_button

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+          "vmType": "Standard_D4s_v3",
+          "vmCount": 4,
+          "autoScale": true
+        },
+        "disposeOnDeletion": true
+      }
+    }
+  }
+}
+```
```
pinin4fjords, line 294: `tw compute-envs export` (0.32.0) does not emit a `computeEnv`/`config` wrapper. The real output is a flat object keyed by `discriminator`, with the config fields hoisted to the top level. I ran this against an Azure Batch Forge CE and got the shape below. The actual file also carries default/null fields (`deleteJobsOnCompletion`, `managedIdentityClientId`, `labels`, ...) that aren't worth showing here. The `forge` block survives, so the prose still lands.

````suggestion
```json title="forge-ce.json" hl_lines="2 7"
{
  "discriminator": "azure-batch",
  "region": "eastus",
  "workDir": "az://work",
  "waveEnabled": true,
  "fusion2Enabled": true,
  "forge": {
    "headPool": { "vmType": "Standard_D2s_v3", "vmCount": 1, "autoScale": true },
    "workerPool": { "vmType": "Standard_D4s_v3", "vmCount": 4, "autoScale": true },
    "disposeOnDeletion": true,
    "dualPoolConfig": true
  }
}
```
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 25]
```diff
@@ -0,0 +1,739 @@
...
+- `azurerm`: creates a head pool and a worker pool on that account.
+- `seqera`: creates the compute environment pointing at those pools.
+
+The manual marker: `head_pool` is set to the pool Terraform just made and there is **no `forge` block**. That one difference is what makes the compute environment manual instead of Forge. `nextflow_config` routes tasks to the worker pool.
+
+Let's walk the key blocks of `terraform/compute-env/main.tf`.
+
+The first block declares the providers and pins their versions. The `seqera` provider is pinned to exactly `0.40.1`:
+
+```terraform
```
pinin4fjords: Title the excerpt and highlight the pinned `seqera` provider line the prose calls out.

````suggestion
```terraform title="terraform/compute-env/main.tf" hl_lines="5"
````

**Review by pinin4fjords — COMMENTED** [review]
Ran the Platform side of this end to end in a container (terraform 1.9.8, tw 0.32.0, seqerakit 0.5.7, against an Azure Batch workspace). Notes from actually executing it, plus code-block formatting / pedagogy suggestions inline.

**Verified works exactly as written:**
- 2.4 Terraform: create, idempotent re-apply (`No changes`), and the `master`->`v2.4` update-in-place all match the documented console output verbatim. Nice bit of teaching.
- 2.2 `tw pipelines add` second-run "already exists" error is accurate.

**Needs attention:**
- **The `forge-ce.json` block is not what `tw` emits** - real `tw compute-envs export` returns a flat object keyed by `discriminator`, with no `computeEnv`/`config` wrapper. Suggestion inline.
- **Name clashes between the four add methods.** The UI, `tw`, `seqerakit`, and Terraform steps all add the same `rnaseq-nf-$WORKSHOP_USER`. Running the section top to bottom, `seqerakit add` errors on its *first* run because the `tw add` step already created that name, so the "run it again and it errors" point lands for the wrong reason. Suffixing each method's name (e.g. `-tw`, `-sk`, `-tf`) keeps them independent.
- **The Terraform pipeline create can return a 403** depending on the token's workspace role. The same `main.tf` and provider 0.40.1 created the pipeline fine with a Maintain-capable token, so this is environmental rather than a code bug. Worth a troubleshooting note ("403 = your token's role can't create pipelines in this workspace") rather than a code change.
- **`azure-batch-manual` isn't guaranteed for Maintain/Launch-only learners** (they never run section 1), so the hardcoded compute-env name needs to be a placeholder. I ran the quest's commands against an existing CE with a different name to confirm the rest holds.

**Couldn't exercise:** 1.1/1.3/1.4 (Azure provisioning + cloud console), the UI steps, and real launches (didn't fire runs).

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 26]
```diff
@@ -0,0 +1,739 @@
...
+    seqera  = { source = "seqeralabs/seqera", version = "0.40.1" }
+    azurerm = { source = "hashicorp/azurerm", version = ">= 3.0" }
+    random  = { source = "hashicorp/random", version = ">= 3.0" }
+  }
+}
+```
+
+Next we configure the providers. The `seqera` provider reads `TOWER_ACCESS_TOKEN` from the environment, so there is no token argument here:
+
+```terraform
```
pinin4fjords:
````suggestion
```terraform title="terraform/compute-env/main.tf"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 27]
```diff
@@ -0,0 +1,739 @@
...
+}
+
+provider "seqera" {
+  server_url = var.server_url
+}
+```
+
+Next we can define some variables we can use throughout our deployment. In this case, we will set the node sizes and details once and reuse them throughout the configuration:
+
+```terraform
```
pinin4fjords: Highlight the VM sizes, since the surrounding text is about node sizing.

````suggestion
```terraform title="terraform/compute-env/main.tf" hl_lines="2 3"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 28]
```diff
@@ -0,0 +1,739 @@
...
+  # Azure VM size names embed that count (Standard_E16ds_v5 -> 16), so derive it
+  # from the size rather than hardcoding a number that can drift out of sync.
+  head_vm_cores   = tonumber(regex("^Standard_[A-Z]+([0-9]+)", local.head_vm_size)[0])
+  worker_vm_cores = tonumber(regex("^Standard_[A-Z]+([0-9]+)", local.worker_vm_size)[0])
+}
+```
+
+Next we can refer to resources that already exist. These are called `data` in Terraform. Here we refer to the Azure Batch account:
+
+```terraform
```
pinin4fjords:
````suggestion
```terraform title="terraform/compute-env/main.tf"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 29]
```diff
@@ -0,0 +1,739 @@
...
+# Existing Azure resources the customer owns. Referenced, not managed.
+data "azurerm_batch_account" "existing" {
+  name                = var.batch_account_name
+  resource_group_name = var.batch_account_rg
+}
+```
+
+Now the resources Terraform does create. The pool names must stay unique across re-creates, so we first generate a short random suffix to append to them:
+
+```terraform
```
pinin4fjords:
````suggestion
```terraform title="terraform/compute-env/main.tf"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 30]
```diff
@@ -0,0 +1,739 @@
...
+resource "random_string" "suffix" {
+  length  = 6
+  special = false
+  upper   = false
+}
+```
+
+Then the node pools themselves. Terraform creates two pools in Azure Batch with the `azurerm_batch_pool` resource. You can add as many as you like, but the details must match the values the Azure provider expects. Here we start a fixed-size head pool and a dynamically sized worker pool:
+
+```terraform
```
pinin4fjords:
````suggestion
```terraform title="terraform/compute-env/main.tf"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 31]
```diff
@@ -0,0 +1,739 @@
...
+    ignore_changes = [auto_scale, fixed_scale]
+  }
+}
+```
+
+Each pool in `main.tf` also carries a `start_task` (elided above) that runs once per node at startup. Forge adds an equivalent task to the pools it creates; manual pools get nothing, so we replicate the essential parts: install `azcopy` (Nextflow's Azure Batch executor uses it to stage data) and register the AppArmor profile Fusion needs on Ubuntu 24.04 nodes. See the `node_start_task_script` local in the file for the full script.
+
+Once we've built the Azure Batch node pools, we add the Seqera Platform compute environment that points to them. The compute environment needs Azure credentials to talk to your cloud. We added those to the workspace earlier (see Prerequisites), so we look them up by name rather than storing keys here. The `seqera_credentials` data source lists the credentials in the workspace, and we pick out the one whose name matches:
+
+```terraform
```
pinin4fjords:
````suggestion
```terraform title="terraform/compute-env/main.tf"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 32]
```diff
@@ -0,0 +1,739 @@
...
+  azure_credentials_id = one([
+    for c in data.seqera_credentials.all.credentials : c.id
+    if c.name == var.azure_credential_name
+  ])
+}
+```
+
+Finally, the compute environment itself. This is the manual marker in code: `head_pool` points at the pool we just created and `nextflow_config` routes tasks to the worker pool's queue. `enable_wave` and `enable_fusion` turn on Wave (container provisioning) and the Fusion file system, the same checkboxes you would tick in the UI; with Fusion enabled, Wave must be too. Because it references `azurerm_batch_pool.head.name` and `azurerm_batch_pool.worker.name`, Terraform knows to create the pools first:
+
+```terraform
```
pinin4fjords: This is the single most important highlight in the lesson: `head_pool` set + no `forge` block is the "manual" marker the prose keeps referring to. Highlight `head_pool` and `nextflow_config` so the reader's eye lands on it.

````suggestion
```terraform title="terraform/compute-env/main.tf" hl_lines="14 17"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 33]
```diff
@@ -0,0 +1,739 @@
...
+        nextflow_config = "process.queue = '${azurerm_batch_pool.worker.name}'\n"
+      }
+    }
+  }
+}
+```
+
+The last block is the output the next tier needs:
+
+```terraform
```
pinin4fjords:
````suggestion
```terraform title="terraform/compute-env/main.tf"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 34]
```diff
@@ -0,0 +1,739 @@
...
+
+```bash
+tw pipelines delete --name="rnaseq-nf-$WORKSHOP_USER"
+```
+
+### 2.3. Add a pipeline with `seqerakit`
+
+`seqerakit` is a wrapper over `tw` that reads a YAML file and runs the underlying `tw` commands. It keeps the configuration as code, so a teammate can reproduce the exact same pipeline. `seqerakit/add-rnaseq.yml` describes the pipeline:
+
+```yaml
```
pinin4fjords: Title it and highlight the two fields a learner actually edits (the unique name and the compute environment).

````suggestion
```yaml title="seqerakit/add-rnaseq.yml" hl_lines="2 6"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 35]
```diff
@@ -0,0 +1,739 @@
...
+No changes. Your infrastructure matches the configuration.
+```
+
+That is the difference. `tw` and `seqerakit` errored on the second run because they perform an action every time. Terraform manages whether the pipeline **exists**: it is already there, so there is nothing to do. Apply ten times, still one pipeline. This is what keeps you from accumulating competing, half-duplicated resources in a shared workspace. To remove the pipeline, `terraform apply -destroy` with the same vars.
+
+**Update in place.** Terraform does more than create-or-skip; it reconciles the pipeline to whatever the config says. The config tracks the `master` branch, which moves as the pipeline gets new commits. Pin it to a fixed release tag instead, so every launch runs the same code. Edit one line: in `main.tf`, line 35 of the `launch` block, change the revision from the `master` branch to the `v2.4` release tag.
+
+=== "After"
+
+    ```hcl title="main.tf" hl_lines="4" linenums="32"
```
pinin4fjords: `title="main.tf"` is ambiguous (there are two `main.tf` files in this quest). Point at the pipeline one.

````suggestion
    ```hcl title="terraform/pipeline/main.tf" hl_lines="4" linenums="32"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 36]
```diff
@@ -0,0 +1,739 @@
...
+        # The master branch moves; pin a release tag (e.g. v2.4) for reproducible launches.
+        revision       = "v2.4"
+        compute_env_id = var.compute_env_id
+        work_dir       = var.work_dir
+      }
+    ```
+
+=== "Before"
+
+    ```hcl title="main.tf" hl_lines="4" linenums="32"
```
pinin4fjords:
````suggestion
    ```hcl title="terraform/pipeline/main.tf" hl_lines="4" linenums="32"
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 37]
```diff
@@ -0,0 +1,739 @@
...
+- Be comfortable with the command line and basic Nextflow concepts
+
+
+---
+
+## 0. Get started
+
+### Open the training codespace
+
+The Codespace contains all the tools you need (Terraform, `tw`, `seqerakit`); you
```
pinin4fjords: Worth a caveat: "you install nothing yourself" only holds once `ghcr.io/nextflow-io/training:latest` is rebuilt from the updated dev devcontainer (which adds the `terraform`, `seqera-cli`, and `uv-tools` features). The production `.devcontainer/devcontainer.json` pulls the prebuilt image and adds no `features`, so until that image is rebuilt the Codespace has none of these tools. I confirmed today's `:latest` ships without `terraform`/`tw`/`seqerakit`. Probably a merge-order / image-rebuild coordination note.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+2. Click **Add token**, name it (e.g. `platform-automation`), and click **Add**.
+3. Copy the token now. The Platform shows it only once.
+4. In the Codespace terminal, export it under both names:
+
+```bash
+export TOWER_ACCESS_TOKEN=<paste-token>
+export SEQERA_ACCESS_TOKEN=$TOWER_ACCESS_TOKEN
+```
+
+Terraform, `tw`, and `seqerakit` all read the token from these variables. Check it worked:
```
pinin4fjords: `SEQERA_ACCESS_TOKEN` looks unused by anything in this quest. Both `terraform/*/main.tf` headers say the `seqera` provider reads `TOWER_ACCESS_TOKEN` from the env, and `tw`/`seqerakit` use `TOWER_ACCESS_TOKEN` too. Either drop `SEQERA_ACCESS_TOKEN`, or state which tool needs it. As written, "all read the token from these variables" isn't quite accurate.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+tw info
+```
+
+This prints the API endpoint and the authenticated user. If it errors, the token is wrong or not exported.
+
+!!! warning
+
+    This only exists in a single terminal session, if you open a new terminal you will need to export them again.
+
+For Seqera Enterprise, also set `SEQERA_API_URL` (and the `server_url` Terraform variable) to your install's API URL. Everything else is identical.
```
pinin4fjords: Double-check the variable name here: `tw` (and therefore `seqerakit`) read the Enterprise endpoint from `TOWER_API_ENDPOINT`, not `SEQERA_API_URL`. The Terraform `server_url` variable is correct. Worth being precise so Enterprise users don't set a var nothing reads.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+
+The same compute environment can be created with increasing control over the cloud. We build it two ways.
+
+### 1.1. Click-ops a compute environment with Batch Forge
+
+Batch Forge is the hands-off option: the Platform reaches into your cloud and creates the resources for you. Most convenient, least control.
+
+In the side bar, open **Compute Environments** and click **Create compute environment**. Give it a name, pick your cloud platform, and select the credentials for that cloud. The rest of the form differs by provider; recommended settings:
+
+**AWS**
```
pinin4fjords: 1.1 walks both AWS and Azure, but everything after it (the 1.2 export example, the 1.3 Terraform, and the `az://work` work-dir throughout 2-3) is Azure-only. An AWS learner who follows 1.1 hits a wall at 1.3 with no path forward. Either say up front "the rest of this quest assumes Azure Batch" or note where AWS users diverge.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
+# Seqera Platform Automation
+
+The Seqera Platform does not run your work. It is an API and a control plane over your cloud. It hands jobs to a compute environment, the compute environment runs them on cloud VMs, and the Platform reads back state, logs, and exit codes.
```
pinin4fjords: Repo convention is one sentence per line (cleaner diffs). This paragraph (and most prose in the file) packs several per line. Not blocking, but worth a reflow pass. Example:

```suggestion
The Seqera Platform does not run your work.
It is an API and a control plane over your cloud.
It hands jobs to a compute environment, the compute environment runs them on cloud VMs, and the Platform reads back state, logs, and exit codes.
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+cd /workspaces/training/side-quests/platform_automation/seqerakit
+export COMPUTE_ENVIRONMENT=<compute environment name>
+
+seqerakit launch-rnaseq.yml --dryrun   # prints the tw command, changes nothing
+seqerakit launch-rnaseq.yml            # launches for real
+```
+
+The dry run shows the underlying `tw` command. Run it twice and you get two runs. That is the imperative model: do the thing, now.
+
+One advantage of using Seqerakit is the YAML forms a template of actions to perform. Because of this we can do two more things: firstly, we can save it and re-use it later. Secondly, we can launch the same pipeline several times by adding more launch blocks to the YAML file. This is useful if you want to launch the same pipeline with different parameters or on different compute environments. Let's try and launch the pipeline now with
```
pinin4fjords: `Seqerakit` should be `` `seqerakit` `` (lowercase, backticked) to match the rest of the doc, and this is a long run-on. Suggest splitting.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+| ------------ | ------------- | --------------------------- |
+| Manages      | existence     | actions                     |
+| Run twice    | no-op         | two runs                    |
+| Mental model | desired state | do the thing, now           |
+
+- **Roles stratify what each token can do.** A Maintain token defines pipelines and Actions; a Launch token can only trigger an Action and nothing else. The Action is the safe handoff from maintainers to launchers (and to automation).
+
+## What's next?
+
+- The AI half of the workshop is the [Building with Seqera AI (CoScientist)](../co_scientist/index.md) side quest. It uses the same `rnaseq-nf` pipeline and API endpoints, but drives them via AI agents.
```
pinin4fjords: Heads up that this relative link 404s until #963 merges (the PR description notes it). If #968 can merge first, consider temporarily pointing at the side-quests index, or gating merge order, so `master` never ships a broken link.

pinin4fjords (commented):
A second pass, this time on teaching style, lining the quest up with the conventions in the other side quests (`essential_scripting_patterns`, `metadata`, `working_with_files`, ...). The framing (learning goals, role table) and the wrap-up (per-section Takeaways, Summary, the declarative-vs-imperative table) already match the house style well, and the idempotency demonstrations (run `tw add` twice, apply Terraform twice) are exactly the "run and observe" loop those quests are built on. These suggestions push the rest of the body toward the same "do, then explain" rhythm.

- **Collapsible command output.** The other quests wrap expected output in `??? success`/`??? failure "Command output"` so the page stays scannable and learners self-check. Converted the four bare `console` blocks inline.
- **Make the hand-off artifact visible from the CLI.** 1.2/1.4 read the compute environment back over the API; mirrored that for the Maintain tier so the Launchpad pipeline is something the learner *sees*, not just reads about.
- **Less passive code-walking in 1.3.** Reframed the "walk every block" lead-in inline. The fuller restructure (too big for a clean suggestion): once `terraform.tfvars` is filled, run `terraform init` + `terraform plan` up front so learners see the two pools and the CE the config will create, then explain only the manual-marker block against that plan and drop the providers/locals/pools into a `??? abstract` for reference. No other quest narrates eight blocks before the first run.
- **Voice.** Minor nudge from tour-guide "we build" to second-person "you'll build".

**One diagram (recommendation, no code to suggest):** the Admin -> Maintain -> Launch artifact hand-off and the Platform -> Batch -> VM submission relationship are the quest's core mental model and are inherently spatial. A single `.excalidraw.svg` (per the repo's diagram convention) would anchor the whole quest; the other quests use images where the idea is spatial.

The CE-name placeholder edit is already covered by existing threads plus the `hl_lines` highlight in my earlier review, so I haven't duplicated it here.

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+
+Admin users create compute environments and assign roles to everyone else. If you only have Maintain, an Admin on your team hands you a `compute_env_id`; start at section 2. If you only have Launch, they hand you a Launchpad pipeline (and optionally an Action) to run; start at section 3. Each section opens with what it requires.
+
+---
+
+## 1. Admin: compute environments and cloud
+
+**Requires:** Owner or Admin role, and permissions in the cloud environment. If you are using a Cloud provider, authenticate with them first via their CLI (e.g. `aws sso login`).
+
+The same compute environment can be created with increasing control over the cloud. We build it two ways.
```
pinin4fjords: Small voice nudge toward the second person the other quests use ("you'll build", not "we build").

```suggestion
The same compute environment can be created with increasing control over the cloud. You'll build it two ways.
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+Terraform manages cloud resources declaratively: you describe what should exist and Terraform makes it so. It does not run your work; it only creates the compute environment. `side-quests/platform_automation/terraform/compute-env/` does it in a single `main.tf`, two providers in one apply:
+
+- data sources: the existing Batch account and the Azure credentials already
+  stored in the Platform, referenced, not created.
+- `azurerm`: creates a head pool and a worker pool on that account.
+- `seqera`: creates the compute environment pointing at those pools.
+
+The manual marker: `head_pool` is set to the pool Terraform just made and there is **no `forge` block**. That one difference is what makes the compute environment manual instead of Forge. `nextflow_config` routes tasks to the worker pool.
+
+Let's walk the key blocks of `terraform/compute-env/main.tf`.
```
pinin4fjords: The other quests never narrate eight code blocks before the learner runs anything; they run early and explain against real output. Reframe the lead-in so the providers/locals/pools read as reference and the focus lands on the one block that encodes the lesson. (See the review summary for the fuller plan-first restructure: `init` + `plan` once `terraform.tfvars` is filled, then collapse the boilerplate into a `??? abstract`.)

```suggestion
The whole file is in the repo to read at your own pace, and most of it is standard Terraform: the providers, the locals, and the two Azure Batch pools. Only one block encodes the lesson that makes this compute environment _manual_, so we'll focus there and treat the rest as reference.
```

docs/en/docs/side_quests/platform_automation/index.md
```diff
@@ -0,0 +1,739 @@
...
+```bash
+cd /workspaces/training/side-quests/platform_automation/seqerakit
+seqerakit add-rnaseq.yml
+```
+
+If you run it again, it errors, the same way `tw` did, because the pipeline already exists:
+
+```console
+The pipelines resource already exists and will not be created. Please set 'on_exists: overwrite' to replace the resource or set 'on_exists: ignore' to ignore this error.
+```
```
pinin4fjords: The other quests wrap expected output in collapsible `??? success`/`??? failure "Command output"` blocks so the page stays scannable and learners can self-check. This one is an error, so `failure`.

````suggestion
??? failure "Command output"

    ```console
    The pipelines resource already exists and will not be created. Please set 'on_exists: overwrite' to replace the resource or set 'on_exists: ignore' to ignore this error.
    ```
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 47]
```diff
@@ -0,0 +1,739 @@
...
+
+Open the Launchpad: `rnaseq-nf-$WORKSHOP_USER` is there, with no run. Now apply again:
+
+```bash
+terraform apply -var="username=$WORKSHOP_USER" -var="workspace_id=<ID>" -var="compute_env_id=<CE_ID>"
+```
+
+```console
+No changes. Your infrastructure matches the configuration.
+```
```
pinin4fjords: Collapsible output, matching the house style.

````suggestion
??? success "Command output"

    ```console
    No changes. Your infrastructure matches the configuration.
    ```
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 48]
```diff
@@ -0,0 +1,739 @@
...
+  # seqera_pipeline.rnaseq_nf will be updated in-place
+  ~ resource "seqera_pipeline" "rnaseq_nf" {
+      ~ launch = {
+          ~ revision = "master" -> "v2.4"
+            # (4 unchanged attributes hidden)
+        }
+    }
+
+Plan: 0 to add, 1 to change, 0 to destroy.
+```
```
pinin4fjords: Collapsible output, matching the house style.

````suggestion
??? success "Command output"

    ```console
    Terraform will perform the following actions:

      # seqera_pipeline.rnaseq_nf will be updated in-place
      ~ resource "seqera_pipeline" "rnaseq_nf" {
          ~ launch = {
              ~ revision = "master" -> "v2.4"
                # (4 unchanged attributes hidden)
            }
        }

    Plan: 0 to add, 1 to change, 0 to destroy.
    ```
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 49]
```diff
@@ -0,0 +1,739 @@
...
+tw launch rnaseq-nf-$WORKSHOP_USER
+```
+
+It should show something like:
+
+```console
+  Workflow 2ZXaU1AzEn7Onk submitted at [Organization / Workspace] workspace.
+
+    https://cloud.seqera.io/orgs/Organization/workspaces/Workspace/watch/2ZXaU1AzEn7Onk
+```
```
pinin4fjords: Collapsible output, matching the house style.

````suggestion
??? success "Command output"

    ```console
      Workflow 2ZXaU1AzEn7Onk submitted at [Organization / Workspace] workspace.

        https://cloud.seqera.io/orgs/Organization/workspaces/Workspace/watch/2ZXaU1AzEn7Onk
    ```
````

File: docs/en/docs/side_quests/platform_automation/index.md [hunk 50]
```diff
@@ -0,0 +1,739 @@
...
+cd /workspaces/training/side-quests/platform_automation/terraform/pipeline
+terraform init
+terraform apply -var="username=$WORKSHOP_USER" -var="workspace_id=<ID>" -var="compute_env_id=<CE_ID>"
+```
+
+!!! tip
+
+    Copy `terraform.tfvars.example` to `terraform.tfvars` and fill it in so you stop passing `-var` flags.
+
+Open the Launchpad: `rnaseq-nf-$WORKSHOP_USER` is there, with no run. Now apply again:
```
pinin4fjords: Section 1.2/1.4 read the compute environment back over the API; mirror that here so the learner *watches* the Launchpad pipeline appear (the artifact handed to the Launch tier), instead of only being told it's there.

````suggestion
Confirm it landed, in the UI (open the Launchpad) or from the terminal, the same way you read the compute environment back in section 1.2:

```bash
tw pipelines list
```

`rnaseq-nf-$WORKSHOP_USER` is there, with no run. Now apply again:
````

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: 19c3285804bdac982e70257648f643cf4dfdadf8

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: One was needed for Seqera CLI but we aren't using it here. let's remove.

e532fe6c2ccd15f74c58677615fa1f9ff72e8e15

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: f4c4132c446f8aeafc21eeb6b109e7d48f733376

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: I wasn't sure about this.

Credentials are complex and cause many problems. I'd rather leave them to docs if possible and, for the purposes of a tutorial, assume they exist. I'll add a quick sentence saying they have to exist

adamrtalbot: 1b406116384d6f2d5e68094500cc9822a7592052

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: These were about as tight as I could get the values before the setting depends on a bunch of stuff in your own cloud account 😢

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: Platform tries to validate the credentials but fails because it defaults to westeurope, but our Batch account is not in that region. I've raised a bug but it won't be fixed soon.

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: Hmmmm do we want to encourage people to save them to a file? I'd prefer them to lose them once or twice than save them and expose it by accident.

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: Agreed. Snipped right down and told users to open the Terraform file themselves if they want to explore it. Suggested the key points that it refers to variables and resources from the Terraform.

adamrtalbot: Also simplified the heck out of it.

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: you mean the subscription ID and stuff? Good point, I'll add some quick bullet points.

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: This is strictly for admins who have access to their own cloud portal. In a live training I might show access to the relevant resources with a screen share.

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: e880543e9ad64b23f2792f6152c95c17e8d17af8

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: e880543e9ad64b23f2792f6152c95c17e8d17af8

docs/en/docs/side_quests/platform_automation/index.md
adamrtalbot: It's hard coded in the Terraform.

I left it because the Terraform spins up a few hot nodes to work on 😉 so the training will go much faster.

It probably needs a comment to acknowledge your compute env name might be different 🤔

PR #970, opened 2026-06-23
Migrate hello_nf-core to nf-core tools 4.0.2 + v2 parser

pinin4fjords:
Modernizes the **hello_nf-core** course from nf-core tools **3.5.2 → 4.0.2** so it runs under the Nextflow 26.04 **default v2 parser**, removing the `NXF_SYNTAX_PARSER=v1` workaround.

## What changed

- **Solutions** regenerated against the nf-core **4.0.2** template + **nf-schema 2.7.2**; all 5 (`core-hello-start`, `part2`-`part5`) verified running to completion under the v2 parser on native x86.
- **`validate_params`**: the `--validate_params false` CLI flag is a no-op under v2 (bare CLI values are typed as strings; `"false"` is truthy - nf-core/tools#4100). Parts 2-4 now set `validate_params = false` in config (a real boolean), preserving the **Part 5** lesson that teaches the schema-validation transformation. The CLI flag is gone from all commands.
- **Docs** (Parts 1-5) regenerated from genuine 4.0.2/v2 runs, including the new `nf-core/tools version 4.0.2` create banner and the `Generated container configs...` line.
- **Orientation** rewritten: drops the v1 instruction, documents that the course runs under the v2 default, and adds guidance that **boolean params must be set via `-params-file`/config, not bare CLI flags** (the v2 string-typing behavior - it also affected Part 1's `--skip_trim`, now shown via a params file).
- **Devcontainer** `uv-tools` pin → `nf-core==4.0.2`.

## Known 4.0 behavior documented

`nf-core modules install` raises `NotADirectoryError` on the course's single-file local modules (4.0 expects directory-style); the install still completes, so Part 3 now verifies via `modules.json` with a warning admonition.

## Not verified (inherent)

Interactive nf-core TUIs (`pipelines create`, `schema build`, `modules create`) can't be driven headlessly; their banners/versions are genuine 4.0.2, the click-through steps are unchanged.


vdauwera (approved):
Made some additional refinements and added a note of caution about potentially needing to set syntax parser to v1 to run other nf-core pipelines

PR #979, opened 2026-09-08
run-tutorial: delegate walkthrough execution, verify fixes independently

pinin4fjords:
## Summary

Updates the `run-tutorial` skill to run more agentically, inspired by the coordinator/delegate/verify-independently pattern in [seqeralabs/nf-metro's `fix-issue` skill](https://github.com/seqeralabs/nf-metro/tree/main/.claude/skills/fix-issue):

- Adds a **Delegation Pattern** section: Phase 2/2B per-lesson execution is delegated to a fresh subagent that returns a structured report, keeping raw command transcripts out of the coordinator's context instead of inlining every `nextflow run` in one long session.
- Adds a **mandatory independent verification gate**: a proposed fix must be confirmed by a subagent that did not diagnose it, before it reaches the PR workflow. This catches a fix that only suppresses the symptom (e.g. loosening a snippet or deleting a failing check) rather than actually correcting the tutorial.
- Renumbers `references/pr-workflow.md` to make the verification gate step 1, ahead of categorizing and presenting fixes.

## Changes

- `.claude/skills/run-tutorial/SKILL.md`: new "Delegation Pattern" section; Phase 4 now requires the independent verification gate before proceeding.
- `.claude/skills/run-tutorial/references/pr-workflow.md`: new step 1 (Independent Verification), remaining steps renumbered.

## Testing

Skill/prompt documentation change only; no code to run. Reviewed the resulting SKILL.md and reference file for correct step numbering and working relative links.

Generated with [Claude Code](https://claude.ai/code)


pinin4fjords:
Addressed both blocking findings:

1. **Hash-check regex** (pre-existing bug, not introduced by this PR, but caught here since it's in the section this PR references): the invalid-hex grep only checked the character immediately after `[`, so `[ab/12g456]` slipped through while `[j6/abc123]` didn't. Fixed in both `SKILL.md` and `references/acceptable-differences.md` (the command was duplicated in both) to validate the full hash pattern.
2. **Independent verifier evidence**: the verifier now also receives the original discrepancy and the observed output/file contents that exposed it, plus an instruction to reproduce the smallest relevant check rather than just judge the diff plausible. Its report must state verdict, evidence checked, and remaining uncertainty. Also added a two-strike bound on the diagnose-verify loop (escalate to the user on a second failure) and a note to treat subagent-supplied lesson/output content as data, not instructions.

Left the verifier report schema and prompt-injection framing lightweight rather than fully formalizing them - this is a docs-authoring skill, not the nf-metro engineering harness it's inspired by, so I kept the gate proportionate to that scope.

vdauwera (approved):
This is fantastic, going to be very helpful for the wide-ranging updates like what I'm doing now.

PR #980, opened 2026-09-08
Fix pedagogical gaps and content drift in workflows_of_workflows side quest

pinin4fjords:
## Summary

- Explains that `GREETING_WORKFLOW.out.timestamped` is intentionally left off `publish:` once `main.nf` composes the greeting workflow, since it becomes `TRANSFORM_WORKFLOW`'s input in section 2 rather than a standalone output like it was in the section 1.1 demo run. Tightens the section-1 takeaway bullet that previously implied both channels were still published.
- Clarifies that leftover files from the standalone section 1.1/2.1 runs remain in `results/` root alongside the new subdirectories shown in the 1.3.3/2.3.3 output listings, and why that's expected (section 2.1 depends on those leftovers as its input).
- Replaces the unresolved `<!-- TODO -->` in the assignment overview with a concrete 5-step data-flow list.
- Adds the required `#!/usr/bin/env nextflow` shebang to all six entry-workflow scripts (starter and solution), updating the matching doc snippets and `hl_lines` counts.
- Adds a missing `nextflow.config` to the starter directory so it publishes real file copies instead of symlinks, matching the solution directory and every other side quest.
- Style cleanup: removes all em dashes and casual "Let's" phrasing, splits several two-sentences-per-line paragraphs, fixes a stray ` ```nextflow ` fence, and normalizes the "Key patterns" list's bold/colon convention.

## Test plan

- [x] `.github/check_headings.py` passes with 0 errors
- [x] `/check-highlights` re-verified all 12 `hl_lines` blocks after the shebang-driven line shifts
- [x] Solution pipeline runs clean in Docker (`nextflow run main.nf`, exit 0), `results/` contains real file copies (not symlinks) with the new starter config
- [x] Starter-tree progression re-verified: section 1.1's standalone `greeting.nf` run still produces the files section 2.1's standalone `transform.nf` run depends on

### DIFF
--- a/docs/en/docs/side_quests/workflows_of_workflows/index.md
+++ b/docs/en/docs/side_quests/workflows_of_workflows/index.md
@@ -1,8 +1,10 @@
 # Workflows of Workflows
 
-When you're developing a pipeline, you often find yourself creating similar sequences of processes for different data types or analysis steps. You might end up copying and pasting these process sequences, leading to duplicated code that's hard to maintain; or you might create one massive workflow that's difficult to understand and modify.
+When you're developing a pipeline, you often find yourself creating similar sequences of processes for different data types or analysis steps.
+You might end up copying and pasting these process sequences, leading to duplicated code that's hard to maintain; or you might create one massive workflow that's difficult to understand and modify.
 
-One of the most powerful features of Nextflow is its ability to compose complex pipelines from smaller, reusable workflow modules. This modular approach makes pipelines easier to develop, test, and maintain.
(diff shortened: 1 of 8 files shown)


vdauwera:
Not specific to this branch but fyi I'm planning to update the style guide to not take out 'casual "Let's" phrasing' because in my experience more casual phrasing is beneficial to the learning experience.

vdauwera (approved):
Looks great, this addresses a few things that bugged me last time I taught this so thank you for fixing!

PR #981, opened 2026-09-09
Fix variable naming and simplify output renaming in metadata side quest

pinin4fjords:
## Summary

Fixes found via an agentic run-tutorial walkthrough of the metadata side quest (`docs/en/docs/side_quests/metadata/`), plus two issues flagged by manual review of the current published page.

## Changes

| Location | Before | After | Reason |
|----------|--------|-------|--------|
| Throughout (first introduced §1.3) | `COWPY`'s recording-file variable alternated between `input_file` and `recording` | Consistently `recording` from its first appearance | The name changed direction more than once (`input_file` → `recording` at §1.4.1 → back to `input_file` at §1.5.3, staying that way through §2.4/§3.2), undercutting a section whose point is teaching a consistent interface |
| Lines 10, 743, 751 | Em dashes | Rewritten with periods/parentheses | House style: no em dashes |
| §2.4.1 (`modules/cowpy.nf`) | Process renamed its own output file using `${meta.lang}-${input_file}` | Process just emits `tuple val(meta), path("cowpy-${recording}")`, no language-aware renaming | The process doesn't need to know `lang`/`lang_group` exist; per the doc's own convention (§1.5), that kind of customization belongs downstream |
| §2.4.3 (`main.nf` output block) | `path { meta, file -> meta.lang_group }` (directory routing only) | `path { meta, file -> file >> "${meta.lang_group}/${meta.lang}-${file.name}" }` | The `>>` operator lets the output block do both subdirectory routing and renaming in one place, entirely outside the process |
| §2.4.4, §3.1.1, §3.1.2, Summary | Console/tree examples and key-patterns snippets reflected the old process-side renaming | Updated to match the new published filenames (e.g. `germanic/de-cowpy-guten_tag.txt`) and the failure-mode commands (`cowpy-bonjour.txt`, no language prefix, since renaming happens after the process would have already failed) | Keep prose and examples in sync with the restructured code |

Same set of changes applied to the matching solution snapshots under `side-quests/solutions/metadata/{1.3,1.5,2.1,2.2,2.3,2.4,3.2}/` and the starter `side-quests/metadata/modules/cowpy.nf`.

## Testing

- [x] Full tutorial walkthrough (progressive build, sections 0-3.2) run end-to-end in Docker against Nextflow 26.04.4 with the strict v2 parser; all commands and outputs matched the docs before this fix (no pre-existing functional bugs).
- [x] `/validate` (heading numbering, admonitions, `hl_lines`, inline-code formatting) and `/test-example` (fresh run + `-resume` cache-hit) passed clean.
- [x] The `>>`-operator restructure was independently re-verified in a separate session that did not author the fix: reproduced the full §2.4.4 pipeline run, confirmed the `results/` tree matches the doc exactly, confirmed `-resume` still gets full cache hits on both processes, and reproduced both §3.1.1/§3.1.2 failure modes to confirm the updated "Command executed" lines are accurate.
- [x] `uv run .github/check_headings.py` and `prettier --check` both pass.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/en/docs/side_quests/metadata/index.md
+++ b/docs/en/docs/side_quests/metadata/index.md
@@ -7,7 +7,8 @@ This extra information is what we call metadata.
 Metadata is data describing other data.
 Metadata tracks important details about files and experimental conditions, and helps tailor analyses to each dataset's unique characteristics.
 
-Think of it like a library catalog: while books contain the actual content (raw data), the catalog cards provide essential information about each book—when it was published, who wrote it, where to find it (metadata).
+Think of it like a library catalog: while books contain the actual content (raw data), the catalog cards provide essential information about each book.
+That information includes when it was published, who wrote it, and where to find it (metadata).
 In Nextflow pipelines, metadata can be used to:
 
 - Track file-specific information throughout the workflow
@@ -354,15 +355,15 @@ process COWPY {
     container 'community.wave.seqera.io/library/cowpy:1.1.5--3db457ae1977a273'
 
     input:
-    path input_file
+    path recording
     val character
 
     output:
-    path "cowpy-${input_file}"
+    path "cowpy-${recording}"
 
     script:
     """
-    cat ${input_file} | cowpy -c ${character} > cowpy-${input_file}
+    cat ${recording} | cowpy -c ${character} > cowpy-${recording}
     """
 }
 ```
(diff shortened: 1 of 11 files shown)


robsyme (approved):
Fantastic. Definitely big improvements. Approved.

pinin4fjords:
> Fantastic. Definitely big improvements. Approved.

Thanks Rob!

PR #982, opened 2026-09-13
Guard training skills against Nextflow agent-mode console output

vdauwera:
Nextflow 26.04+ detects the CLAUDECODE environment variable (set by Claude Code in every shell it runs) and silently switches to a machine-readable "agent mode" console format ([PIPELINE], [PROCESS], [SUCCESS]/[FAILED] tags) instead of the classic human-facing output learners actually see. Any skill that runs `nextflow` directly and copies its output into docs was at risk of baking in a format no learner will ever encounter (already found once, in docs/en/docs/execution_config/02_resources_and_retries.md).

- Document the issue and the required override (env -u CLAUDECODE NXF_AGENT_MODE=false) in shared/repo-conventions.md
- Apply the override in test-example and run-tutorial's concrete nextflow invocations
- Add a hard-fail check for agent-mode tags to run-tutorial/references/acceptable-differences.md
- Note in docker-setup why containerized runs are unaffected today, and warn against changing that


vdauwera:
@pinin4fjords Found a fun issue-- Claude started complaining that the console output had changed dramatically and we needed to update all the console output text in all the courses. I freaked out a little before realizing it was just Nextflow recognizing Claude and setting agent mode automatically.

